PR #148, opened 2021-10-15
Use broadcast for xsimd>=8

jtilly:
Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,12 +7,20 @@
 Changelog
 =========
 
+3.0.3 - 2021-10-15
+------------------
+
+**Bug fix**
+
+- We now support ``xsimd>=8`` and support alternative jemalloc installations.
+
+
 3.0.2 - 2021-10-14
 ------------------
 
 **Bug fix**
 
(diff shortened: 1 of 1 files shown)


jtilly:
With bdf7fb40ad905c07fb557a67eb2bf518c48ce836, this should also work with `jemalloc` instead of `jemalloc-local`. I'm verifying this [here](https://github.com/Quantco/tabmat/commit/239303e978fb95ee6e648e9181026bf6af3bccb3).

xhochy:
> Also, @xhochy: is it correct that you can't install `tabmat` in an environment where `jemalloc` is installed? Because then [I get](https://github.com/Quantco/tabmat/runs/3908295646?check_suite_focus=true#step:4:129) `- package jemalloc-local-5.2.1-h9c3ff4c_6 has constraint jemalloc <0a0 conflicting with jemalloc-4.4.0-0`

No, you can. You simply cannot combine `jemalloc` and `jemalloc-local` in an environment. For the runtime side, `tabmat` only requires `libjemalloc-local` which can live alongside `jemalloc`.

PR #161, opened 2021-12-01
Use long long for indexing in sparse_sandwich

### LINKED ISSUE #160: BUG: segfault when fitting a GeneralizedLinearRegressor
Requirements: `pip install libsvmdata`

The following script gives me a segfault:
```python
from libsvmdata import fetch_libsvm
from glum import GeneralizedLinearRegressor

X, y = fetch_libsvm("rcv1.binary")
clf = GeneralizedLinearRegressor(alpha=0.01, fit_intercept=False,
                                 family="gaussian")
clf.fit(X, y)
```

Output:
```python
In [1]: %run glum_segfault.py
Dataset: rcv1.binary
[1]    271745 segmentation fault (core dumped)  ipython
```

I'm using glum 2.0.3  

@qb3

jtilly:
Closes #160 

I added a unit test, but it's of not much use, because we don't have enough memory in CI to run it, so I'm skipping it.


src/tabmat/ext/sparse.pyx
```diff
@@ -35,7 +35,7 @@ def sparse_sandwich(A, AT, floating[:] d, win_integral[:] rows, win_integral[:]
 
     cdef floating* dp = &d[0]
 
-    cdef win_integral m = cols.shape[0]
+    cdef long long m = cols.shape[0]
```
xhochy, line 38: Please use the integer types with explicit size. You will need to import `from libc.stdint cimport int64_t` at the top.

```suggestion
    cdef int64_t m = cols.shape[0]
```

xhochy (changes requested):

src/tabmat/ext/sparse.pyx
jtilly: Should this be changed elsewhere as well? Currently, we use `win_integral` (fused type of integral and long long) very liberally throughout (see https://github.com/Quantco/tabmat/pull/87).

xhochy: Yes, it would make more sense if the fused type would also use only sized integers. I guess though that here `int64_t` is the right one always.

MarcAntoineSchmidtQC:
@jtilly, can you wait before merging this. I'm trying to figure out if we can replace `win_integral` with `int64_t` everywhere. I think this might solve future problems and be more efficient computationally.

MarcAntoineSchmidtQC (approved):
I investigated a couple of options but none of them lead to a significant improvement. I've changed a few `win_integral` with `int64_t` which should make the code a bit more robust to very large datasets. The rest looks good to me. We can merge this.

PR #165, opened 2022-01-03
Macos test failing on main

MarcAntoineSchmidtQC:
Somehow the CI for the macos test when pushed to main was not updated at the same time that the same "normal" tests were updated. The culprit was an old version of mamba that solved the environment using pypy instead of cpython.

Checklist
* [ ] ~~Added a `CHANGELOG.rst` entry~~ (only affects the CI so no changelog)

### DIFF
--- a/.github/workflows/conda-build-macos.yml
+++ b/.github/workflows/conda-build-macos.yml
@@ -27,7 +27,7 @@ jobs:
       - uses: conda-incubator/setup-miniconda@35d1405e78aa3f784fe3ce9a2eb378d5eeb62169
         with:
           miniforge-variant: Mambaforge
-          miniforge-version: 4.10.0-0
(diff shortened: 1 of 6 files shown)


jtilly (approved):
Good catch!

There are few more instances of that version of miniforge in the repo:
```
$ git grep 4.10.0-0
.github/workflows/conda-build-macos.yml:30:          miniforge-version: 4.10.0-0
.github/workflows/conda-build-win.yml:27:          miniforge-version: 4.10.0-0
.github/workflows/tests-macos-main.yml:26:          miniforge-version: 4.10.0-0
.github/workflows/tests-win-main.yml:29:          miniforge-version: 4.10.0-0
.github/workflows/tests-win.yml:31:          miniforge-version: 4.10.0-0
```
Maybe update those, too?

PR #167, opened 2022-01-03
Test against nightly builds

jtilly:
This adds a daily CI job to our test suite that test against nightlies of pandas, scipy, and numpy (linux only).

Same as https://github.com/Quantco/glum/pull/502


MarcAntoineSchmidtQC (approved):
Thanks.

PR #168, opened 2022-01-07
implementation of `CategoricalMatrix` with drop_first

### LINKED ISSUE #75: Enable dropping one column from a CategoricalMatrix?
Currently, CategoricalMatrix does not provide an easy way to drop a column. We are required to include a category for every row in the dataset, but in an unregularized setting, it is nice to sometimes drop one column. 

Something sort of like this is already implemented by the `cols` parameter to the matrix vector and sandwich functions.

MarcAntoineSchmidtQC:
closes #75.

Description
------------

This P.R. adds the functionality to "drop" the first level of a `CategoricalMatrix`. With unregularized models having also an intercept, this is necessary to maintain a full rank design matrix.

I decided to always drop the first level. A user could always reorder the categories manually to have the correct reference level.

Future work?
-------------

Some operations could be optimized. For example, all the `_cross_sandwich` operations are done by simply computing the result without dropping a level, and then to drop the first row/column or the resulting output matrix. I would merge this in as-is since I think changing this will significantly complexify the codebase with a minimal improvement in performance.

Still an open question: how do we make this feature available in the `from_pandas` function and in glum?

Tests
-----

I've modified the tests so that one of the `CategoricalMatrix` component of the testing suite drops a level. I've also added a parametrization to most of the tests in `test_categorical_matrix.py` to check for this new feature. I didn't create any new test.

Checklist

### DIFF
--- a/src/tabmat/categorical_matrix.py
+++ b/src/tabmat/categorical_matrix.py
@@ -167,7 +167,14 @@ import numpy as np
 import pandas as pd
 from scipy import sparse as sps
 
-from .ext.categorical import matvec, sandwich_categorical, transpose_matvec
+from .ext.categorical import (
+    matvec,
+    matvec_drop_first,
+    sandwich_categorical,
+    sandwich_categorical_drop_first,
+    transpose_matvec,
+    transpose_matvec_drop_first,
+)
 from .ext.split import sandwich_cat_cat, sandwich_cat_dense
 from .matrix_base import MatrixBase
 from .sparse_matrix import SparseMatrix
@@ -212,12 +219,18 @@ class CategoricalMatrix(MatrixBase):
     cat_vec:
         array-like vector of categorical data.
 
+    drop_first:
+        drop the first level of the dummy encoding. This allows a CategoricalMatrix
+        to be used in an unregularized setting.
+
     dtype:
+        data type
     """
 
     def __init__(
         self,
         cat_vec: Union[List, np.ndarray, pd.Categorical],
+        drop_first: bool = False,
         dtype: np.dtype = np.float64,
     ):
         if pd.isnull(cat_vec).any():
@@ -228,7 +241,8 @@ class CategoricalMatrix(MatrixBase):
         else:
             self.cat = pd.Categorical(cat_vec)
 
-        self.shape = (len(self.cat), len(self.cat.categories))
+        self.drop_first = drop_first
+        self.shape = (len(self.cat), len(self.cat.categories) - int(drop_first))
         self.indices = self.cat.codes.astype(np.int32)
         self.x_csc: Optional[Tuple[Optional[np.ndarray], np.ndarray, np.ndarray]] = None
         self.dtype = np.dtype(dtype)
@@ -299,7 +313,13 @@ class CategoricalMatrix(MatrixBase):
         if out is None:
             out = np.zeros(self.shape[0], dtype=other_m.dtype)
 
-        matvec(self.indices, other_m, self.shape[0], cols, self.shape[1], out)
+        if self.drop_first:
+            matvec_drop_first(
+                self.indices, other_m, self.shape[0], cols, self.shape[1], out
+            )
+        else:
+            matvec(self.indices, other_m, self.shape[0], cols, self.shape[1], out)
+
         if is_int:
             return out.astype(int)
         return out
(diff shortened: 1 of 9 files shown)


tbenthompson:
Drive by PR review. My PR review is simple: 👍 ❤️ 🐶 💯 🎉 🌮

src/tabmat/dense_matrix.py
```diff
@@ -92,7 +92,7 @@ def _matvec_helper(
         vec: Union[List, np.ndarray],
         rows: Optional[np.ndarray],
         cols: Optional[np.ndarray],
-        out: Optional[Union[np.ndarray]],
+        out: Optional[np.ndarray],
```
MarcAntoineSchmidtQC, line 95: This is unrelated to this PR but the linter was warning about this.

src/tabmat/categorical_matrix.py
```diff
@@ -422,6 +458,11 @@ def tocsr(self) -> sps.csr_matrix:
         # TODO: data should be uint8
         data = np.ones(self.shape[0], dtype=int)
 
+        if self.drop_first:
+            return sps.csr_matrix(
+                (data, self.indices, np.arange(self.shape[0] + 1, dtype=int)),
+                shape=(self.shape[0], self.shape[1] + 1),
```
lbittarello, line 464:
```suggestion
                shape=(self.shape[0], self.shape[1] + self.drop_first),
```
and collapse the two blocks?

src/tabmat/categorical_matrix.py
```diff
@@ -540,6 +599,16 @@ def multiply(self, other) -> sps.csr_matrix:
...
 
+        if self.drop_first:
+            return sps.csr_matrix(
+                (
+                    np.squeeze(other),
+                    self.indices,
+                    np.arange(self.shape[0] + 1, dtype=int),
+                ),
+                shape=(self.shape[0], self.shape[1] + 1),
+            )[:, 1:]
```
lbittarello, line 610:
```suggestion
                shape=(self.shape[0], self.shape[1] + self.drop_first),
            )[:, self.drop_first :]
```
and collapse the two blocks?

lbittarello (commented):
> how do we make this feature available in the `from_pandas` function and in glum?

Adding a boolean `drop_first` keyword seems easy enough. It'd be trickier to allow users to set it at a column level, but I'm not sure if there's a use case.

src/tabmat/categorical_matrix.py
MarcAntoineSchmidtQC: This was done on purpose. If you look closely, the `drop_first` block also has a slicing operation at the end. This is inefficient because it always copies the data, even if you select everything. Therefore, only the `drop_first` version has this inefficiency.

src/tabmat/categorical_matrix.py
MarcAntoineSchmidtQC: same thing here. If we can avoid slicing it would be ideal, so I keep them separate.

jtilly:
This is super cool! Thanks for working on this.

> Adding a boolean drop_first keyword seems easy enough. 

Yes, let's just add a `drop_first` keyword to `from_pandas` and then pass it through to both
https://github.com/Quantco/tabmat/blob/f6747ad203f1cc857efdfeecdc2a525d27517a8f/src/tabmat/constructor.py#L77-L79
and
https://github.com/Quantco/tabmat/blob/f6747ad203f1cc857efdfeecdc2a525d27517a8f/src/tabmat/constructor.py#L97

That would make `drop_first` in `CategoricalMatrix` consistent with the behavior in [pd.get_dummies](https://pandas.pydata.org/docs/reference/api/pandas.get_dummies.html).

As for the glum implementation: Can't we just add `drop_first` to the init of [`GeneralizedLinearRegressorBase`](https://github.com/Quantco/glum/blob/fd38196a2ebbbfa375a329ecd9138877a000baa5/src/glum/_glm.py#L645-L703) and then pass that argument along? Passing it along to [here](https://github.com/Quantco/glum/blob/fd38196a2ebbbfa375a329ecd9138877a000baa5/src/glum/_glm.py#L89) is straightforward. There are a few things that require care (e.g. [this](https://github.com/Quantco/glum/blob/fd38196a2ebbbfa375a329ecd9138877a000baa5/src/glum/_glm.py#L1698-L1706) and [this](https://github.com/Quantco/glum/blob/fd38196a2ebbbfa375a329ecd9138877a000baa5/src/glum/_glm.py#L1708-L1738)), but this should all be easy to fix.

lbittarello (approved):
🚀

PR #172, opened 2022-03-01
Multiply method

MarcAntoineSchmidtQC:
This PR adds a `multiply` method ~~and the `__mul__` and `__rmul__` special methods~~ to the matrices. This is needed for [this glum PR](https://github.com/Quantco/glum/pull/526).

Checklist

### DIFF
--- a/src/tabmat/categorical_matrix.py
+++ b/src/tabmat/categorical_matrix.py
@@ -607,7 +607,7 @@ class CategoricalMatrix(MatrixBase):
     def multiply(self, other) -> SparseMatrix:
         """Element-wise multiplication.
 
-        This assumes that ``other`` is a vector of size self.shape[1].
+        This assumes that ``other`` is a vector of size self.shape[0].
         """
         if self.shape[0] != other.shape[0]:
             raise ValueError(
--- a/src/tabmat/categorical_matrix.py
+++ b/src/tabmat/categorical_matrix.py
@@ -607,7 +607,7 @@ class CategoricalMatrix(MatrixBase):
     def multiply(self, other) -> SparseMatrix:
         """Element-wise multiplication.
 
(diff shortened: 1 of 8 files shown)


src/tabmat/categorical_matrix.py
```diff
@@ -597,38 +597,46 @@ def _cross_sparse(
...
 
         res = term_1.T.dot(_row_col_indexing(other, rows, R_cols)).A
         return res
 
-    def multiply(self, other) -> sps.csr_matrix:
-        """Element-wise multiplication of each column with other."""
+    def multiply(self, other) -> SparseMatrix:
+        """Element-wise multiplication.
+
+        This assumes that ``other`` is a vector of size self.shape[1].
```
KrzysztofChomiakQC, line 610: This line of docstring seems inconsistent with line 612 below, where the check is performed with ``self.shape[0]`` not on ``self.shape[1]``

MarcAntoineSchmidtQC: Thanks! 🤦

DavidMeesterQC:
Thanks! Looks good to me 🚀

src/tabmat/categorical_matrix.py
```diff
@@ -597,38 +597,46 @@ def _cross_sparse(
...
 
         res = term_1.T.dot(_row_col_indexing(other, rows, R_cols)).A
         return res
 
-    def multiply(self, other) -> sps.csr_matrix:
-        """Element-wise multiplication of each column with other."""
+    def multiply(self, other) -> SparseMatrix:
+        """Element-wise multiplication.
+
+        This assumes that ``other`` is a vector of size self.shape[0].
```
lbittarello, line 610:
```suggestion
        This assumes that ``other`` is a vector of size ``self.shape[0]``.
```
?

jtilly (approved):

PR #173, opened 2022-03-04
readthedocs reconfiguration

MarcAntoineSchmidtQC:
See #170 and similar PR in glum: https://github.com/Quantco/glum/pull/520

Checklist
* [ ] ~~Added a `CHANGELOG.rst` entry~~


jtilly (approved):

environment.yml
```diff
@@ -3,6 +3,7 @@ channels:
   - conda-forge
   - nodefaults
 dependencies:
+  - python<3.10
```
xhochy, line 6: What is the issue with 3.10? Can you add a comment about that here?

MarcAntoineSchmidtQC: Seems that the incompatibility between python 3.10, sphinx, and readthedocs has been fixed. I re-ran PR #170 and it was successful. No need for this PR anymore. glum is also passing with python 3.10 so I will be changing it there as well.

PR #192, opened 2022-06-30
Don't use march=native when building wheel

jtilly:
I think this should help address https://github.com/Quantco/glum/issues/558


MarcAntoineSchmidtQC:
Failing CI is unrelated to this PR.

MarcAntoineSchmidtQC (approved):

PR #201, opened 2022-08-24
Pin xsimd to < 9

MarcAntoineSchmidtQC:
xsimd v9.0 is the reason for the nightly failure (renamed `hadd` to `reduce_add`). I fixed the nightly failure but it turns out that xsimd v9.0 seems to be incompatible with macos, and all of our tests on macos are failing.

Checklist

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -7,7 +7,7 @@ on:
     - '.github/workflows/daily.yml'
     - '.github/workflows/test.sh'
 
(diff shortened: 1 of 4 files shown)


xhochy (approved):
Can you please report the macos problem upstream with xsimd?

MarcAntoineSchmidtQC:
Here it is: https://github.com/xtensor-stack/xsimd/issues/810

PR #206, opened 2022-10-14
Build for all possible linux architectures

### LINKED ISSUE #205: Build script in PyPI source version uses default `jemalloc`
I see the build script for linux uses `jemalloc` with `disable-tls`:
[  "./autogen.sh --disable-cxx --with-jemalloc-prefix=local --with-install-suffix=local --disable-tls --disable-initial-exec-tls",](https://github.com/Quantco/tabmat/blob/ec8fcd237ed5f581fa61efe91733117366343d24/pyproject.toml#L73)

However, the source distribution in PyPI doesn't run that script when installing it through `pip`, relying instead on whatever `jemalloc` it finds when it tries to compile. If, for example, one tries to install `tabmat` from source through `pip`, it will later on fail to import, complaining about an error with `jemalloc`:
```
cannot allocate memory in static TLS block
```

xhochy:
Fixes #205

Checklist

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.10.2
         env:
-          CIBW_ARCHS_LINUX: auto aarch64 ppc64le s390x
+          CIBW_ARCHS_LINUX: auto aarch64
 
       - uses: actions/upload-artifact@v3
         with:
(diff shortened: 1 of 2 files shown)


MarcAntoineSchmidtQC:
@xhochy, the cibuildwheel options are currently set in pyproject.toml. It really makes sense to put it where you've put it because it needs the emulation, but it leads to the config being scattered.

If you don't mind, I can push on your branch to add a comment in the toml file saying that the `archs_linux` option is set in the CI workflow file.

xhochy:
Feel free to push. I haven't seen on how to set this otherwise.

MarcAntoineSchmidtQC (approved):

PR #218, opened 2023-01-26
Build wheels for Apple Silicon

xhochy:
Checklist


xhochy:
I'm confused that the `Build and upload to PyPI` workflow isn't triggered 😕

MarcAntoineSchmidtQC:
Okay, I've looked into this today. Seems that the arm64 wheels are actually compiled from an x86_64 and this is causing a bunch of issues. I was able to make it work with a lot of fixes:

- Do not use homebrew to install dependencies. Seems really complicated to install an arm64 package on an x86_64 machine. I used conda instead and specified `-c conda-forge/osx-arm64`.
- To work with openmp, we used to install llvm using homebrew. This didn't work for me. It always tried to use the x86_64 version of openmp. Instead I used the apple default clang, removed the `-fopenmp` flag, and added the -lomp flag (copied what sklearn is doing). OpenMP is installed using conda from the previous point.
- I added this `-Wl,-rpath,$CONDA/envs/build/lib` to the LDFLAGS envvar. I have no idea what it does but without it the "delocate" step fails. I copied this from sklearn as well.

I will clean up my test branch and create a PR into your branch.

PR #221, opened 2023-01-31
Use jemalloc-local instead of jemalloc

### LINKED ISSUE #220: dlopen symbol not found issue with M1 wheel
I'm getting the following error when trying to import tabmat using the `tabmat-3.1.3-cp38-cp38-macosx_11_0_arm64.whl` wheel:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/sidharthkapur/test-workspace/tabmat-test/lib/python3.8/site-packages/tabmat/__init__.py", line 1, in <module>
    from .categorical_matrix import CategoricalMatrix
  File "/Users/sidharthkapur/test-workspace/tabmat-test/lib/python3.8/site-packages/tabmat/categorical_matrix.py", line 182, in <module>
    from .sparse_matrix import SparseMatrix
  File "/Users/sidharthkapur/test-workspace/tabmat-test/lib/python3.8/site-packages/tabmat/sparse_matrix.py", line 6, in <module>
    from .ext.sparse import (
ImportError: dlopen(/Users/sidharthkapur/test-workspace/tabmat-test/lib/python3.8/site-packages/tabmat/ext/sparse.cpython-38-darwin.so, 0x0002): symbol not found in flat namespace '_je_aligned_alloc'
```

Steps to reproduce:
```sh
python3.8 -m venv ./tabmat-test
source ./tabmat-test/bin/activate
pip install tabmat==3.1.3
# Collecting tabmat
#   Using cached tabmat-3.1.3-cp38-cp38-macosx_11_0_arm64.whl (590 kB)
# Collecting pandas
#   Downloading pandas-1.5.3-cp38-cp38-macosx_11_0_arm64.whl (10.8 MB)
#      |████████████████████████████████| 10.8 MB 15.3 MB/s
# Collecting numpy
#   Using cached numpy-1.24.1-cp38-cp38-macosx_11_0_arm64.whl (13.8 MB)
# Collecting scipy
#   Using cached scipy-1.10.0-cp38-cp38-macosx_12_0_arm64.whl (28.8 MB)
# Collecting python-dateutil>=2.8.1
#   Using cached python_dateutil-2.8.2-py2.py3-none-any.whl (247 kB)
# Collecting pytz>=2020.1
#   Downloading pytz-2022.7.1-py2.py3-none-any.whl (499 kB)
#      |████████████████████████████████| 499 kB 82.4 MB/s
# Collecting six>=1.5
#   Using cached six-1.16.0-py2.py3-none-any.whl (11 kB)
# Installing collected packages: six, pytz, python-dateutil, numpy, scipy, pandas, tabmat
# Successfully installed numpy-1.24.1 pandas-1.5.3 python-dateutil-2.8.2 pytz-2022.7.1 scipy-1.10.0 six-1.16.0 tabmat-3.1.3
python -c 'import tabmat'
```

xhochy:
Fixes #220

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -1,7 +1,6 @@
 name: Build and upload to PyPI
 
 on:
-  push:
   pull_request:
   release:
     types:
--- a/build_tools/prepare_macos_wheel.sh
+++ b/build_tools/prepare_macos_wheel.sh
@@ -9,4 +9,4 @@ else
     CONDA_CHANNEL="conda-forge/osx-64"
 fi
 
-conda create -n build -c $CONDA_CHANNEL jemalloc xsimd llvm-openmp
+conda create -n build -c $CONDA_CHANNEL jemalloc-local xsimd llvm-openmp
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ LDFLAGS="-Wl,-rpath,$CONDA/envs/build/lib -L$CONDA/envs/build/lib -lomp"
 CFLAGS="-I$CONDA/envs/build/include"
 CXX="/usr/bin/clang++"
 CC="/usr/bin/clang"
+JE_INSTALL_SUFFIX="local"
 
 [tool.cibuildwheel.windows]
 before-all = [
--- a/setup.py
+++ b/setup.py
@@ -57,12 +57,13 @@ if sys.platform == "win32":
     # make sure we can find xsimd headers
     include_dirs.append(os.path.join(sys.prefix, "Library", "include"))
 elif sys.platform == "darwin":
-    jemalloc_config = shutil.which("jemalloc-config")
-    if jemalloc_config is None:
+    if "JE_INSTALL_SUFFIX" in os.environ:
+        je_install_suffix = os.environ["JE_INSTALL_SUFFIX"]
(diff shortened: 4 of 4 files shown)


MarcAntoineSchmidtQC:
Sorry, this is my fault. I faced this issue and used jemalloc instead of jemalloc-local instead, just to try, and forgot to fix the issue later.

Looking at the problem, I think the issue is that we are installing jemalloc inside a conda environment, but we are not activating the environment (because we want to build the wheel with right python executable). Therefore, the call [here](https://github.com/Quantco/tabmat/blob/0bf12140dba8ffd9974a4cfc96a0859a53b6297f/setup.py#L60) returns nothing (since jemalloc is not on the path).

To be honest, I'm not sure I understand why we needed jemalloc local in the first place. I'm really not sure how to best solve this. Adding the "bin" folder of the conda environment is probably not going to work (we could simply activate the environment instead). Making setup.py work with this by adding an extra environment variable adds extra complexity on top of something already complex.

MarcAntoineSchmidtQC:
The environment variable path might not be too crazy. Setting something like JE_INSTALL_SUFFIX and then only going through the whole `shutil.which("jemalloc-config")` if it doesn't exists sounds decent. We can then add this environment variable to the wheel building CI manually. What do you think @xhochy?

xhochy:
Sensible approach!

PR #235, opened 2023-03-27
Use 64bit ints indexing for F-contiguous case

### LINKED ISSUE #234: Sandwich product fails for large F-contiguous matrices
For large F-contiguous matrices, we still run into a seg fault due to an overflow issue [here](https://github.com/Quantco/tabmat/blob/main/src/tabmat/ext/dense_helpers-tmpl.cpp#L237). 

Minimal example: 

```python
import numpy as np 
from tabmat import DenseMatrix 

ii32 = np.iinfo(np.int32) 
K = 1000 
N = ii32.max//(K-1) + 1 # to make sure we overflow 
rng = np.random.default_rng(seed=12345) 
X =DenseMatrix(
np.ndarray(buffer=rng.random(N * K), dtype=float, shape=[N, K], order="F")) 

y = rng.random(N) 

X.sandwich(y)
```

ChristianRoerigQC:
This fixes #234.
 
Checklist


CHANGELOG.rst
```diff
@@ -7,6 +7,13 @@
 Changelog
 =========
 
+3.1.6 - 2022-03-XX
```
jtilly, line 10:
```suggestion
3.1.6 - 2022-03-27
```

jtilly (approved):

MarcAntoineSchmidtQC (approved):
🚀  Great! Thanks @ChristianRoerigQC

PR #240, opened 2023-03-28
Use 64bit int for indexing in cross sandwich product

### LINKED ISSUE #239: Cross sandwich product fails for split matrices with large dense matrix part
Unfortunately, we still run into a seg fault for the split matrix case. Apologies, that my prior minimal example only considered the pure dense matrix case. Increasing the size of the iterator variables i and j, [here](https://github.com/Quantco/tabmat/blob/main/src/tabmat/ext/sparse_helpers-tmpl.cpp#L71-#L73), fixes it. 

Split matrix minimal example:

```python
import numpy as np 
from tabmat import DenseMatrix, SparseMatrix, SplitMatrix 
ii32 = np.iinfo(np.int32) 

K = 1000 
N = ii32.max//(K-1) + 1 # to make sure we overflow for K * N 
rng = np.random.default_rng(seed=12345) 

X_dense = DenseMatrix( 
            np.ndarray( 
                buffer=rng.random(N * K), 
                dtype=float, 
                shape=[N, K]) 
            ) 

X_sparse = SparseMatrix(np.random.choice([0,1], size=[N, 50]), dtype=float) 

X_split= SplitMatrix([X_dense, X_sparse]) 

y = rng.random(N) 

X_split.sandwich(y) 


```

ChristianRoerigQC:
This fixes #239.

Please double check that I haven't missed more cases so that we tackle any final overflow issues in this PR. 

Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@
 Changelog
 =========
 
+3.1.7 - 2022-03-28
+------------------
+
+**Bug fix:**
+
+- We fixed a bug in the cross sandwich product, which would previously segfault for very large matrices.
+
+
 3.1.6 - 2022-03-27
 ------------------
 
--- a/src/tabmat/ext/sparse_helpers-tmpl.cpp
+++ b/src/tabmat/ext/sparse_helpers-tmpl.cpp
@@ -6,6 +6,12 @@
 
 #include "alloc.h"
 
+#ifdef _WIN32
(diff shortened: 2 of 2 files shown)


MarcAntoineSchmidtQC (approved):
As discussed, this PR will probably be modified by #241, which covers more instances where we should be using 64-bits integers.

But we will merge this PR to main to make a quick release.

PR #241, opened 2023-03-28
C++ types refactoring

MarcAntoineSchmidtQC:
This PR refactors the C++/cython types used. Here are the design decisions I made:

- To deal with windows integers, I am templating all functions to use template types Int and F. Currently, a lot of functions are using F as the template float type, but are hard-coding “int”. I am replacing all the “int” with “Int”.
- Importantly, I switched (again, sorry) our usage of size_t (and now SIZE_T), to always using Py_ssize_t. This is the recommended type by cython to deal with indexing. It is signed, so windows should not complain. See this [SO thread](https://stackoverflow.com/questions/20987390/cython-why-when-is-it-preferable-to-use-py-ssize-t-for-indexing)
- I am doing the type casting when needed. A lot of our functions deal with array indices. The segfault problem we have is that the index can be stored in a 32 bits integer, but when we multiply two indices, we get an overflow error. To solve this without sacrificing a lot of performance, I’m casting the indices to Py_ssize_t only when doing arithmetics on them. see example below.
- When we define new variables in C++ (that are not passed from the caller), and those variable represent indices, I am always using Py_ssize_t. **This is the decision I'm the least confident is optimal - feedback appreciated**. Another option would be to use Int (the same type as what the caller sends), and casting to Py_ssize_t only when necessary.

## Performance analysis

Only showing the dense + 2 categoricals here but the others are qualitatively the same. This is on an M1 Pro.
Old:
image
New:
image

So the performance seems to be in the same ballpark. Maybe there's a very tiny slowdown due to using 64-bit ints for a lot of indexing, but it's clear to me that even if this wasn't randomness, it wouldn't be enough to negate the benefits of using 64-bit indices.

TODO:

- [x] Analyze potential performance hit from this change

Checklist

### DIFF
--- a/src/tabmat/ext/dense_helpers-tmpl.cpp
+++ b/src/tabmat/ext/dense_helpers-tmpl.cpp
@@ -151,17 +151,17 @@ void dense_base${kparallel}(F* R, F* L, F* d, F* out,
                 Py_ssize_t kmin, Py_ssize_t kmax, Int innerblock, Int kstep) 
 {
     constexpr std::size_t simd_size = xsimd::simd_type<F>::size;
-    for (Int imin = imin2; imin < imax2; imin+=innerblock) {
-        Int imax = imin + innerblock; 
+    for (Py_ssize_t imin = imin2; imin < imax2; imin+=innerblock) {
+        Py_ssize_t imax = imin + innerblock; 
         if (imax > imax2) {
             imax = imax2; 
         }
-        for (Int jmin = jmin2; jmin < jmax2; jmin+=innerblock) {
-            Int jmax = jmin + innerblock; 
+        for (Py_ssize_t jmin = jmin2; jmin < jmax2; jmin+=innerblock) {
+            Py_ssize_t jmax = jmin + innerblock; 
             if (jmax > jmax2) {
                 jmax = jmax2; 
             }
-            Int i = imin;
+            Py_ssize_t i = imin;
             % for IBLOCK in [4, 2, 1]:
             {
                 ${outer_i(kparallel, IBLOCK, [4, 2, 1])}
--- a/src/tabmat/ext/dense_helpers-tmpl.cpp
+++ b/src/tabmat/ext/dense_helpers-tmpl.cpp
@@ -24,12 +24,6 @@
 
 #include "alloc.h"
 
-#ifdef _WIN32
-    #define SIZE_T long long
-#else
-    #define SIZE_T size_t
-#endif
-
 #if XSIMD_VERSION_MAJOR >= 8
     #define XSIMD_BROADCAST broadcast
 #else
@@ -149,25 +143,25 @@ namespace xs = xsimd;
 </%def>
 
 <%def name="dense_base_tmpl(kparallel)">
-template <typename F>
+template <typename Int, typename F>
 void dense_base${kparallel}(F* R, F* L, F* d, F* out,
-                int out_m,
-                int imin2, int imax2,
-                int jmin2, int jmax2, 
-                int kmin, int kmax, int innerblock, int kstep) 
+                Py_ssize_t out_m,
+                Py_ssize_t imin2, Py_ssize_t imax2,
+                Py_ssize_t jmin2, Py_ssize_t jmax2, 
+                Py_ssize_t kmin, Py_ssize_t kmax, Int innerblock, Int kstep) 
 {
     constexpr std::size_t simd_size = xsimd::simd_type<F>::size;
-    for (int imin = imin2; imin < imax2; imin+=innerblock) {
-        int imax = imin + innerblock; 
+    for (Int imin = imin2; imin < imax2; imin+=innerblock) {
+        Int imax = imin + innerblock; 
         if (imax > imax2) {
             imax = imax2; 
         }
-        for (int jmin = jmin2; jmin < jmax2; jmin+=innerblock) {
-            int jmax = jmin + innerblock; 
+        for (Int jmin = jmin2; jmin < jmax2; jmin+=innerblock) {
+            Int jmax = jmin + innerblock; 
             if (jmax > jmax2) {
                 jmax = jmax2; 
             }
-            int i = imin;
+            Int i = imin;
             % for IBLOCK in [4, 2, 1]:
             {
                 ${outer_i(kparallel, IBLOCK, [4, 2, 1])}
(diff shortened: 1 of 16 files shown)


MarcAntoineSchmidtQC:
FYI, @adityagoel4512

src/tabmat/ext/dense_helpers-tmpl.cpp
```diff
@@ -149,25 +143,25 @@ namespace xs = xsimd;
...
-                int kmin, int kmax, int innerblock, int kstep) 
+                Py_ssize_t out_m,
+                Py_ssize_t imin2, Py_ssize_t imax2,
+                Py_ssize_t jmin2, Py_ssize_t jmax2, 
+                Py_ssize_t kmin, Py_ssize_t kmax, Int innerblock, Int kstep) 
 {
     constexpr std::size_t simd_size = xsimd::simd_type<F>::size;
-    for (int imin = imin2; imin < imax2; imin+=innerblock) {
-        int imax = imin + innerblock; 
+    for (Int imin = imin2; imin < imax2; imin+=innerblock) {
```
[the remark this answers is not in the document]
adityagoel4512: If the template arg `Int` is `int32_t` there's an implicit cast anyway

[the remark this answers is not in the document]
MarcAntoineSchmidtQC: Good catch. Originally, imax2 was an Int, but I switched it and forgot to switch the downstream variables.

MarcAntoineSchmidtQC:
Forgot to mention one important thing I did in this PR: I reviewed the dependencies and remove leftovers (libblas and mkl-include). These were necessary at some point but the code pointing to them was removed (a long time ago).

jtilly (approved):

jtilly (approved):

src/tabmat/ext/sparse_helpers-tmpl.cpp
```diff
@@ -33,92 +33,91 @@ void _csr_dense${order}_sandwich(
...
 
-                        Int Cj = Cjj;
-                        Int Cjmax2 = Cjj + ((Cjmax - Cjj) / simd_size) * simd_size;
+                        Py_ssize_t Cj = Cjj;
+                        Py_ssize_t Cjmax2 = Cjj + ((Cjmax - Cjj) / simd_size) * simd_size;
                         for (; Cj < Cjmax2; Cj+=simd_size) {
-                            auto Bsimd = xs::load_aligned(&R[(Ck-Ckk)*jblock+(Cj-Cjj)]);
-                            auto outsimd = xs::load_aligned(&outtemp.get()[Ci*nB_cols_rounded+Cj]);
+                            auto Bsimd = xs::load_aligned(&R[(Py_ssize_t) (Ck-Ckk) * jblock + (Cj-Cjj)]);
+                            auto outsimd = xs::load_aligned(&outtemp.get()[(Py_ssize_t) Ci * nB_cols_rounded + Cj]);
```
adityagoel4512, line 104:
```suggestion
                            auto outsimd = xs::load_aligned(&outtemp.get()[Ci * nB_cols_rounded + Cj]);
```

src/tabmat/ext/sparse_helpers-tmpl.cpp
```diff
@@ -33,92 +33,91 @@ void _csr_dense${order}_sandwich(
...
                             outsimd = xs::fma(Qsimd, Bsimd, outsimd);
-                            outsimd.store_aligned(&outtemp.get()[Ci*nB_cols_rounded+Cj]);
+                            outsimd.store_aligned(&outtemp.get()[(Py_ssize_t) Ci * nB_cols_rounded + Cj]);
```
adityagoel4512, line 106:
```suggestion
                            outsimd.store_aligned(&outtemp.get()[Ci * nB_cols_rounded + Cj]);
```

src/tabmat/ext/sparse_helpers-tmpl.cpp
```diff
@@ -33,92 +33,91 @@ void _csr_dense${order}_sandwich(
...
                         }
 
                         for (; Cj < Cjmax; Cj++) {
-                            outtemp.get()[Ci*nB_cols_rounded+Cj] += Q * R[(Ck-Ckk)*jblock+(Cj-Cjj)];
+                            outtemp.get()[(Py_ssize_t) Ci * nB_cols_rounded + Cj] += Q * R[(Py_ssize_t) (Ck-Ckk) * jblock + (Cj-Cjj)];
```
adityagoel4512, line 110:
```suggestion
                            outtemp.get()[Ci * nB_cols_rounded + Cj] += Q * R[(Py_ssize_t) (Ck-Ckk) * jblock + (Cj-Cjj)];
```

src/tabmat/ext/sparse_helpers-tmpl.cpp
```diff
@@ -33,92 +33,91 @@ void _csr_dense${order}_sandwich(
...
             }
         }
 
-        for (Int Ci = 0; Ci < nA_cols; Ci++) {
-            for (Int Cj = 0; Cj < nB_cols; Cj++) {
+        for (Py_ssize_t Ci = 0; Ci < nA_cols; Ci++) {
+            for (Py_ssize_t Cj = 0; Cj < nB_cols; Cj++) {
                 #pragma omp atomic
-                out[Ci*nB_cols+Cj] += outtemp.get()[Ci*nB_cols_rounded+Cj];
+                out[(Py_ssize_t) Ci * nB_cols + Cj] += outtemp.get()[(Py_ssize_t) Ci * nB_cols_rounded + Cj];
```
adityagoel4512, line 120:
```suggestion
                out[(Py_ssize_t) Ci * nB_cols + Cj] += outtemp.get()[Ci * nB_cols_rounded + Cj];
```

src/tabmat/ext/sparse_helpers-tmpl.cpp
```diff
@@ -33,92 +33,91 @@ void _csr_dense${order}_sandwich(
...
         auto outtemp = make_aligned_unique<F>(
             nA_cols * nB_cols_rounded,
             alignment
         );
-        for (Int Ci = 0; Ci < nA_cols; Ci++) {
-            for (Int Cj = 0; Cj < nB_cols; Cj++) {
-                outtemp.get()[Ci*nB_cols_rounded+Cj] = 0.0;
+        for (Py_ssize_t Ci = 0; Ci < nA_cols; Ci++) {
+            for (Py_ssize_t Cj = 0; Cj < nB_cols; Cj++) {
+                outtemp.get()[(Py_ssize_t) Ci * nB_cols_rounded + Cj] = 0.0;
```
adityagoel4512, line 58:
```suggestion
                outtemp.get()[Ci * nB_cols_rounded + Cj] = 0.0;
```

PR #243, opened 2023-03-29
Move common flags to setup.py

adityagoel4512:


setup.py
```diff
@@ -154,6 +154,14 @@
...
     packages=find_packages(where="src"),
     install_requires=["numpy", "pandas", "scipy"],
-    ext_modules=cythonize(ext_modules, annotate=False),
+    ext_modules=cythonize(
+        ext_modules,
+        annotate=False,
+        compiler_directives={
+            "boundscheck": False,
+            "wraparound": False,
+            "cdivision": True,
```
MarcAntoineSchmidtQC, line 163: I think it would be helpful to add some way to set boundscheck and wraparound to True when debugging. Suggestion: check for the environment variable "TABMAT_DEBUG" when doing pip install. If set, then turn on the boundscheck and the wraparound.

xhochy: No, using O0 will modify (undefined) behavior and thus you won't be able to debug a lot of problems. Thus, don't set it.

setup.py
```diff
@@ -49,6 +49,11 @@
 # add numpy headers
 include_dirs = [np.get_include()]
 
+# check if debug build
+debug_build = "TABMAT_DEBUG" in os.environ
```
MarcAntoineSchmidtQC, line 53: I would prefer checking for a value for this environment variable. I think the usual value would be 1 for true and 0 for false.

```suggestion
debug_build = os.getenv("TABMAT_DEBUG", False)
# then check that the value makes sense (if the value is "false", this would evaluate to true).
```

PR #245, opened 2023-04-09
Use language_level=3 for Cython

jjerphan:
Even though this compiles without any problem, this specifies that the Python 3 syntax is to be used for Cython compilation.

This allows removing some warnings.

---

For now, Cython directives are sitting at the head of Cython sources. Moreover, some functions are documented but do not need to be because of the directives.

Hence, would it make sense to add those directives in the call to `cythonize` here instead?


MarcAntoineSchmidtQC (approved):
Looks good. Will integrate with #243.

Issue #256, opened 2023-06-05
`matvec` inconsistent behavior when used with the `col` argument.


MartinStancsicsQC:
## Description

`matvec` has inconsistent behavior between the various matrix types, and also compared to the docs. According to the docs, `X.matvec(b, cols=cols)` returns the product `X[:, cols] @ b`. In reality, this is not always the case, and the behavior differs between various matrix types.

## Example

```python
import numpy as np
import tabmat as tm

X_dense = tm.DenseMatrix(np.array([
    [1., 2., 3.],
    [4., 5., 6.],
]))
X_split = tm.SplitMatrix([X_dense])
```

Consider the following cases (`X` stands for both `X_dense` and `X_split`):

- `X.matvec([1., 1., 1.])`
  - Expected: `[6, 15]`
  - Got: `[6, 15]` (`DenseMatrix.matvec` returns a `DenseMatrix` instead of a `numpy.ndarray`, but that's probably fine as that is a subclass)

- `X.matvec([1., 1., 1.], cols=[0, 1])`
  - Expected: Error because of shape mismatch
  - Got: `[3, 9]` in both cases
  - Note: It is not necessarily a bug, as `X[:, cols] @ b[cols]` makes a lot of sense, but still inconsistent with the docs

- `X.matvec([1., 1.], cols=[0, 1])`
  - Expected: `[3, 9]`
  - Dense: `[3, 9]`
  - Split: `ValueError: shapes (2, 3) and (2,) not aligned`

- `X.matvec([1.], cols=[0, 1])``
  - Expected: Error because of shape mismatch
  - Dense: `[1, 4]`
  - Split: `ValueError: shapes (2, 3) and (1,) not aligned`

## Environment

Using `tabmat=3.1.7`, conda-forge `osx-arm64` build.

<details>

```
# Name                    Version                   Build  Channel
brotli                    1.0.9                h1a8c8d9_8    conda-forge
brotli-bin                1.0.9                h1a8c8d9_8    conda-forge
bzip2                     1.0.8                h3422bc3_4    conda-forge
ca-certificates           2023.5.7             hf0a4a13_0    conda-forge
certifi                   2023.5.7           pyhd8ed1ab_0    conda-forge
charset-normalizer        3.1.0              pyhd8ed1ab_0    conda-forge
idna                      3.4                pyhd8ed1ab_0    conda-forge
libblas                   3.9.0           17_osxarm64_openblas    conda-forge
[...]

PR #257, opened 2023-06-07
Matvec dimensions

MartinStancsicsQC:
Checklist

Update the documentation to reflect the actual behavior of `matvec` and `matvec_transpose`: `X.matvec(b, cols=cols) == X[:, cols] @ b[cols]`. Also, checks have been added to detect dimension mismatch and raise a `ValueError` in the case of all `MatrixBase` subclasses.

"Fixes" #256


MartinStancsicsQC:
Sure thing, added the test. Which then made me realize that I've missed `StandardizedMatrix`, but it's fixed now.

PR #267, opened 2023-06-20
Matrices from formulas

MatthiasSchmidtblaicherQC:
This draft PR is for allowing early review of the `formula` branch.

Checklist


src/tabmat/constructor.py
```diff
@@ -198,3 +202,35 @@ def from_csc(mat: sps.csc_matrix, threshold=0.1):
...
+        pandas DataFrame to be converted.
+    ensure_full_rank: bool, default False
+        If True, ensure that the matrix has full structural rank by categories.
+    """
+    if isinstance(context, int):
+        if hasattr(sys, "_getframe"):
+            frame = sys._getframe(context + 1)
+            context = LayeredMapping(frame.f_locals, frame.f_globals)
+        else:
+            context = None  # pragma: no cover
```
MatthiasSchmidtblaicherQC, line 230: I don't think we test coverage in tabmat yet, so you can remove pragma annotations,

src/tabmat/constructor.py
```diff
@@ -198,3 +202,35 @@ def from_csc(mat: sps.csc_matrix, threshold=0.1):
...
+    formula: str
+        A formula accepted by formulaic.
+    df: pd.DataFrame
+        pandas DataFrame to be converted.
+    ensure_full_rank: bool, default False
+        If True, ensure that the matrix has full structural rank by categories.
+    """
+    if isinstance(context, int):
+        if hasattr(sys, "_getframe"):
+            frame = sys._getframe(context + 1)
```
MatthiasSchmidtblaicherQC, line 227: `context` is missing in docstring.

src/tabmat/constructor.py
```diff
@@ -198,3 +202,35 @@ def from_csc(mat: sps.csc_matrix, threshold=0.1):
...
+):
+    """
+    Transform a pandas DataFrame to a SplitMatrix using a Wilkinson formula.
+
+    Parameters
+    ----------
+    formula: str
+        A formula accepted by formulaic.
+    df: pd.DataFrame
+        pandas DataFrame to be converted.
```
MatthiasSchmidtblaicherQC, line 221:
```suggestion
        pandas data frame to be converted.
```

src/tabmat/formula.py
```diff
@@ -0,0 +1,313 @@
...
+        return super()._is_categorical(values)
+
+    @override
+    def _check_for_nulls(self, name, values, na_action, drop_rows):
+        if na_action is NAAction.IGNORE:
+            return
+
+        if isinstance(
+            values, dict
+        ):  # pragma: no cover; no formulaic transforms return dictionaries any more
```
MatthiasSchmidtblaicherQC, line 41: As discussed before, we might be able to remove the dictionary support and set a lower pin on `formulaic` in this materializer.

src/tabmat/formula.py
```diff
@@ -0,0 +1,313 @@
...
+    def _combine_columns(self, cols, spec, drop_rows):
+        # Special case no columns
+        if not cols:
+            values = numpy.empty((self.data.shape[0], 0))
+            return SplitMatrix([DenseMatrix(values)])
+
+        # Otherwise, concatenate columns into SplitMatrix
+        return SplitMatrix([col[1].to_non_interactable() for col in cols])
+
+    # Have to override this because of culumn names
```
MatthiasSchmidtblaicherQC, line 90:
```suggestion
    # Have to override this because of column names
```

src/tabmat/formula.py
```diff
@@ -0,0 +1,313 @@
...
+    # (and possibly intercept later on)
```
MatthiasSchmidtblaicherQC, line 91: Which change exactly is not from the parent class in this method?

src/tabmat/formula.py
```diff
@@ -0,0 +1,313 @@
...
+
+    def to_non_interactable(self):
+        if numpy.all(self.multipliers == 1):
+            return CategoricalMatrix(
+                self.cat,
+                drop_first=self.drop_first,
+                dtype=self.dtype,
+            )
+        else:
+            return SparseMatrix(
```
MatthiasSchmidtblaicherQC, line 262: Not sure yet if a sparse matrix is the best way to go here if the multipliers come from a dense matrix. In that case, there should be a constant value for number of categories in `self` under which dense representations should be preferred, if I am not mistaken. But this can be addressed in a future iteration.

src/tabmat/formula.py
```diff
@@ -0,0 +1,313 @@
...
+                spec=spec,
+                drop_rows=drop_rows,
+            ),
+            spec=spec,
+        )
+
+
+class InteractableDenseMatrix(DenseMatrix):
+    def __mul__(self, other):
+        if isinstance(other, (InteractableDenseMatrix, int, float)):
```
MatthiasSchmidtblaicherQC, line 181: How does this square with `DenseMatrix.multiply()` expecting only a vector as input?

src/tabmat/formula.py
```diff
@@ -0,0 +1,313 @@
...
+        return DenseMatrix(self)
+
+    def get_names(self, col):
+        return [col]
+
+
+class InteractableSparseMatrix(SparseMatrix):
+    def __mul__(self, other):
+        if isinstance(other, (InteractableDenseMatrix, InteractableSparseMatrix)):
+            return self.multiply(other)
```
MatthiasSchmidtblaicherQC, line 204: See comment for `InteractableDenseMatrix`

src/tabmat/formula.py
```diff
@@ -0,0 +1,313 @@
...
+
+        if other.drop_first:
+            new_codes -= (cardinality_self - 1)
+            new_codes[new_codes < 0] = 0
+            other_slice = slice(1, None)
+        else:
+            other_slice = slice(None)
+
+        new_categories = [
+            f"{self_cat}:{other_cat}"
```
MatthiasSchmidtblaicherQC, line 286: as discussed before, `__x__` as a separator would fit nicely.

tests/test_formula.py
```diff
@@ -0,0 +1,193 @@
+import formulaic
```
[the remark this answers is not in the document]
MatthiasSchmidtblaicherQC: We should also check if any [base checks ](InteractableCategoricalMatrix) of the materializer would be affected by the `TabmatMaterializer` and add their counterparts here when appropriate.

src/tabmat/formula.py
```diff
@@ -0,0 +1,313 @@
...
+            for other_cat, self_cat in itertools.product(
```
MatthiasSchmidtblaicherQC, line 287: Further idea for a check: compare output to model matrix from a statsmodel (or other GLM package) tutorial with a complicated model matrix (e.g., this [one](https://www.statsmodels.org/dev/examples/notebooks/generated/glm_formula.html)), trying different options like sparse or dense, drop_first or not, full structural rank or not. Doesn't need to be a test because we probably don't want the dependency.

src/tabmat/formula.py
MartinStancsicsQC: Absolutely. There are also a couple other performance improvements that could be implemented in the future.

tests/test_formula.py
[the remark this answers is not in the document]
MartinStancsicsQC: Great idea!

src/tabmat/formula.py
MartinStancsicsQC: [This line](https://github.com/Quantco/tabmat/pull/267/files/ffa4955ed2acb0ab0771dde7144b7370d623ba6f#diff-26ccecbd9eb9394448571abffe87552798612849716b251815fc9151b0773a16R150-R158) (well, line might be a stretch)

src/tabmat/constructor_util.py
```diff
@@ -0,0 +1,32 @@
+from typing import Tuple
```
MatthiasSchmidtblaicherQC, line 1: What is the motivation behind moving this out of `constructor.py`? If we keep it separate, then I would rather organize this in a separate package as:
- `constructor/constructor.py` (current `constructor.py`), 
- `constructor/utils.py` (current `constructor_util.py`), 
- `constructor/__init__.py` (exposes imports from `constructor/constructor.py`).

Then one can still write `from tabmat.constructor import from_pandas` etc. as before, but the constructor is more organized.

.ci_support/linux_aarch64_python3.10_default.____cpython.yaml
```diff
@@ -0,0 +1,16 @@
+BUILD:
```
MatthiasSchmidtblaicherQC, line 1: FYI: diff still shows changes from https://github.com/Quantco/tabmat/commit/e2462cc2791f05a257d9ab7f5268cfe92c0448ea

src/tabmat/formula.py
```diff
@@ -0,0 +1,582 @@
...
+            )
+            (
+                dense_part,
+                sparse_part,
+                dense_idx,
+                sparse_idx,
+            ) = _split_sparse_and_dense_parts(sparse_matrix, sparse_threshold)
+            return SplitMatrix([dense_part, sparse_part], [dense_idx, sparse_idx])
+
+    def get_names(self) -> List[str]:
```
MatthiasSchmidtblaicherQC, line 413: I am a bit confused by the naming convention with `name` and `names`. I see how these two concepts are necessary, but I think that a better naming of `names` can be done :). What about `levels`?

tests/test_formula.py
```diff
@@ -0,0 +1,718 @@
...
+    def data_with_nulls(self):
+        return pd.DataFrame(
+            {"a": [1, 2, None], "A": ["a", None, "c"], "B": ["a", "b", None]}
+        )
+
+    @pytest.fixture
+    def materializer(self, data):
+        return TabmatMaterializer(data)
+
+    @pytest.mark.parametrize("formula,tests", FORMULAIC_TESTS.items())
```
MatthiasSchmidtblaicherQC, line 465: What about moving these tests to a separate file `test_materializer.py` or similar?

MatthiasSchmidtblaicherQC (commented):
Looks great! I will try it out on some examples and report back.

src/tabmat/constructor_util.py
MartinStancsicsQC: The reason is that I ran into a circular import: I need `_split_sparse_dense` in both the `formula.py` and `constructor.py` modules, and we need stuff from `formula.py` in `constructor.py`. If this function is in the latter, then that makes it circular. Your solution sounds good, though. Other options include (if we want to keep the flat structure):

 - Move everything from `formula.py` into `constructor.py` (formulas are related to tabmat construction, anyways)
 - Move the `_split_sparse_dense` to `formula.py` and import in `constructor.py`. In that case we should probably call `formula.py` something else.

src/tabmat/formula.py
MartinStancsicsQC: Yeah, I see your point :) The issue is that the same method name must be used for all `InteractableVector` subclasses, and `InteractableSparseVector` and `InteractableDenseVector` do not really have levels. Maybe `get_column_names`? Or even `get_feature_names`? Or are those that still too ambiguous?

MartinStancsicsQC:
> Looks great! I will try it out on some examples and report back.

Great, thanks for testing it. There was a bug due to a bug in `Splitmatrix.__init__`, which was fixed in 3.1.10. I've now merged main into this branch. Could you please use the most recent commit on the branch when trying it out?

.ci_support/linux_aarch64_python3.10_default.____cpython.yaml
MartinStancsicsQC: I don't see a difference, at least to ef84e7d. Is it still an issue?

src/tabmat/formula.py
MartinStancsicsQC: `tabmat` does not handle two-sided formulas (should it? I can be convinced :)), so I think this test would fit better in `glum` when we implement the formula interface there. We already have `statsmodels` as a dev dependency there so it can even be an automatic unit test.

MatthiasSchmidtblaicherQC: We should focus on the right hand side of the formula. 

I agree with the dependency issue but, on the other hand, formulas are a tabmat functionality, so tests for formula outputs should be in this package. We could test against patsy here and against statsmodels in glum. Or we test only against formulaic and rely on formulaic's tests against patsy to close the loop (I haven't checked that there are any, but I hope that they exist).

Irrespective of whether this becomes a test (and in which library( in the end, it is useful to try these formulas on some data examples. These examples could also become tutorials or examples later on. Here is one based on the link above. It would be great if we could also find one for categorical data and interactions with them (we could use area codes from the insurance example, but very happy to consider other sugggestions).
```
# %%
import tabmat as tm
import statsmodels.api as sm
from patsy import dmatrices

# %%
star98 = sm.datasets.star98.load_pandas().data
star98.dtypes
# %%
formula = "LOWINC + PERASIAN + PERBLACK + PERHISP + PCTCHRT + PCTYRRND + PERMINTE*AVYRSEXP*AVSALK + PERSPENK*PTRATIO*PCTAF"
formula_2s = "NABOVE ~ " + formula
# %%
y, X = dmatrices(formula_2s, data=star98, return_type='dataframe')

# %%
X_tm = tm.from_formula(formula, star98, ensure_full_rank=True).A
assert X.shape == X_tm.shape
# %% patsy seems to shuffle columns and interactions
# Does formulaic have a different convention here or is this an idiosyncracy of the tabmat 
# implementation?
(X - X_tm).head()
# %% model trained with either design matrix should be equivalent
mod = sm.OLS(y, X)
res = mod.fit() 

mod_tm = sm.OLS(y, X_tm)
res_tm = mod_tm.fit() 

assert res.aic == res_tm.aic

# %%
```

.ci_support/linux_aarch64_python3.10_default.____cpython.yaml
MatthiasSchmidtblaicherQC: Now it shows the right diff for me as well...

src/tabmat/formula.py
MartinStancsicsQC: The insurance data would be great for this! It shows off features like converting a categoricals on the fly, mix of numerical and categorical columns, etc.

MartinStancsicsQC: > ```python
> # Does formulaic have a different convention here or is this an idiosyncracy of the tabmat 
> # implementation?
> ```

Yes, that's [formulaic using a different convention](https://matthewwardrop.github.io/formulaic/migration/). `tabmat.from_formula` output should not differ from `formulaic.model_matrix` output in anything other than the return type. There is a test for that [here](https://github.com/Quantco/tabmat/blob/ef84e7dc2e65295b467f8dee0ea84e20ef02b1d2/tests/test_formula.py#L389) and [here](https://github.com/Quantco/tabmat/blob/formula/tests/test_formula.py#L283)

.ci_support/linux_aarch64_python3.10_default.____cpython.yaml
MartinStancsicsQC: Weird. I'll keep my eye open in case something like that pops up again.

MartinStancsicsQC:
An example notebook about the main features of this branch: [https://gist.github.com/MartinStancsicsQC/eaad758d27cb119336804a3d4170aad1](https://gist.github.com/MartinStancsicsQC/eaad758d27cb119336804a3d4170aad1)

CHANGELOG.rst
```diff
@@ -17,7 +24,7 @@ Changelog
 - Fixed ``getcol`` not respecting the ``drop_first`` attribute of a ``CategoricalMatrix``.
 
 3.1.9 - 2023-06-16
-------------------
+-------------------
```
MarcAntoineSchmidtQC, line 27: to remove

CHANGELOG.rst
```diff
@@ -7,6 +7,13 @@
 Changelog
 =========
 
+Unreleased
+----------
+
+**Other changes:**
+
+- Improve the performance of ``from_pandas`` in the case of low-cardinality categorical variables.
```
MarcAntoineSchmidtQC, line 15: Please add a changelog entry for the formula part as well.

MartinStancsicsQC:
> Great work Martin! Only question is a small detail: in the notebook it mentions that if you want to add an intercept you need to specify `+ 0`. I can't load the grammar of formulaic but patsy and others use `+ 1` instead. Is that intentional? Can we use both?

Sorry, that's a typo. We also use `+1` to add the intercept. `+0` or `-1` can be used to remove it explicitly (in line with other libs).

PR #271, opened 2023-06-23
Fix segfault for dense_sandwich

MartinStancsicsQC:
Checklist


MartinStancsicsQC:
I think the deleted cast on L.229 was unnecessary, as `Cjj` is already `Py_ssize_t`, but it would be great is someone could confirm.

jtilly:
The standard CI runners have 7GB of memory, that's probably not enough.

Please add the test to the test suite regardless and [add a marker](https://docs.pytest.org/en/7.1.x/example/markers.html). We won't run the test in CI (make sure to skip the marker [here](https://github.com/Quantco/tabmat/blob/main/.github/workflows/test.sh#L47)), but then it's straightforward to run it locally with `pytest tests`.

MartinStancsicsQC:
Something like this?

CHANGELOG.rst
```diff
@@ -7,6 +7,13 @@
 Changelog
 =========
 
+3.1.10 -- 2023-XX-XX
```
jtilly, line 10:
```suggestion
3.1.10 -- 2023-06-23
```

jtilly (approved):

PR #275, opened 2023-06-29
Improve the performance of `from_pandas` in the case of low-cardinality categoricals

MartinStancsicsQC:
Checklist

When encountering low-cardinality (i.e. `< cat_threshold`) categorical variables, `from_pandas` uses the `pandas.get_dummies` to create a dummy encoding. It turns out that that function is quite slow, and `tabmat` already has a much more performant way of achieving the same result: `CategoricalMatrix.tocsr()`. This PR switches the former with the latter and significantly improves runtime.

## Microbenchmark

```python
import timeit
import numpy as np
import pandas as pd
import tabmat as tm

df = pd.DataFrame({
    "cat": pd.Categorical(np.random.choice(["a", "b", "c"], size=1_000_000)),
})

fun = lambda: tm.from_pandas(df)
timeit.timeit(fun, number=10) / 10 * 1000
```

The implementation on branch `main` takes roughly **140 ms** on my laptop, while the one proposed in this PR finishes in around **12 ms**.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,13 @@
 Changelog
 =========
 
+3.1.11 - 2023-XX-XX
+-------------------
+
+**Other changes:**
+
+- Improve the performance of ``from_pandas`` in the case of low-cardinality categorical variables.
+
 3.1.10 - 2023-06-23
 -------------------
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -7,6 +7,13 @@
 Changelog
 =========
 
+3.1.11 - 2023-XX-XX
```
MarcAntoineSchmidtQC, line 10:
```suggestion
Unreleased
```

MarcAntoineSchmidtQC: I don't really see the benefits of adding the year, or the minor version. Simply Unreleased here will do.

MarcAntoineSchmidtQC (approved):
looks good! Just look at my suggestion for the changelog.

PR #278, opened 2023-07-12
Add column name metadata to `tabmat` matrices

MartinStancsicsQC:
Checklist

## Summary

Add column names and term names metadata to all matrix classes, accessible via the `column_names` and `term_names` properties, or via the `get_names` and `set_names` functions.

## Details

The main idea is that **only DenseMatrix, SparsMatrix and CategoricalMatrix instances have these attributes**, and other types construct theirs on the fly. This reduces complexity and the amount of cases we have to consider. Please see details in comments.

**Column and terms:-** I propose to store store two attributes: column and term names. While it may seem weird from a dataframe perspective, it makes sense when used in a regression setting, especially with a formula interface. A term is an element of a formula (e.g. poly(x) or C(x), while columns are the set of variables they imply. In the case of categoricals, a term is a categorical and a column is one of its dummy columns.

**Names are  optional:** Each name is stored as a vector, can contain None as their elements. When using the column name getter function, those columns will have some dynamically generated name signifying column index (e.g. _col_1 ). These names are not persistent over e.g. matrix concatenation or column reordering. Normal name are.

**Matrix operations:** column names are propagated when concatenating via SplitMatrix or using column indexing.

**Formula interface:** this way of implementing column and term names would streamline the formula materializer a significantly, and a couple hundred lines worth of overloaded functions could be removed.

## Issues

The propagation of these new attributes when using certain methods is inconsistent at best and somewhat incorrect at worst. This inconsistency arises from the fact that dense and sparse matrices are subclassed differently from their respective base classes.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,10 @@ Changelog
 Unreleased
 ----------
 
+**New features:**
+
+- Add column name and term name metadata to ``MatrixBase`` objects. These are automatically populated when initializing a ``MatrixBase`` from a ``pandas.DataFrame``. In addition, they can be accessed and modified via the ``column_names`` and ``term_names`` properties.
+
 **Other changes:**
 
 - Improve the performance of ``from_pandas`` in the case of low-cardinality categorical variables.
--- a/src/tabmat/categorical_matrix.py
+++ b/src/tabmat/categorical_matrix.py
@@ -250,6 +250,9 @@ class CategoricalMatrix(MatrixBase):
         cat_vec: Union[List, np.ndarray, pd.Categorical],
         drop_first: bool = False,
         dtype: np.dtype = np.float64,
+        column_name: Optional[str] = None,
+        term_name: Optional[str] = None,
+        column_name_format: str = "{name}[{category}]",
     ):
         if pd.isnull(cat_vec).any():
             raise ValueError("Categorical data can't have missing values.")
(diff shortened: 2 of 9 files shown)


MartinStancsicsQC:
## Inconsistencies in column name propagation

Take a `DenseMatrix` and a `SparseMatrix` with some names:

```python
>>> import numpy as np
>>> import scipy.sparse as sps
>>> from tabmat import DenseMatrix, SparseMatrix

>>> Xd = DenseMatrix(np.ones((5, 2)), column_names=["d1", "d2"], term_names=["d", "d"])
>>> Xs = SparseMatrix(sps.csc_matrix(np.ones((5, 2))), column_names=["c1", "c2"], term_names=["c", "c"])
```

Indexing correctly propagates column names (explicitly implemented) ✅
```python
>>> Xd[:, [1, 0]].column_names, Xs[:, [1, 0]].column_names
(['d2', 'd1'], ['c2', 'c1']) 
```

Transpose and reshape incorrectly propagates column names for `DenseMatrix` ❌. For `SparseMatrix`, the transpose is not even a `SparseMatrix` (better).
```python
>>> Xd.T.column_names
['d1', 'd2']
```

Other functions correctly propagate column names for `DenseMatrix` ✅, but not for `SparseMatrix` ❌
```
>>> (Xd * 2).column_names, (Xs * 2).column_names
(['d1', 'd2'], [None, None])
```

So ATM this PR is fine for the purposes of getting column names for regressions if the matrices are not really manipulated after creating them, but can lead to some nasty surprises if one is not careful.

MartinStancsicsQC:
I've rebased this branch to the `densematrix-refactor` branch (the proposed refactoring effort to make the API more well-defined and consistent). This solved the issues from my previous comment (inconsistent column name propagation), and column name handling became more explicit and straightforward.

MarcAntoineSchmidtQC (approved):
This is great. Everything looks good to me. Ready to merge to v4 when you want.

MarcAntoineSchmidtQC:
Can you add an entry to changelog please.

PR #281, opened 2023-07-18
Allow missing values in `CategoricalMatrix`

MartinStancsicsQC:
Checklist

This PR adds support for missing values in `CategoricalMatrix`. The default behavior does not change (raise on missing categoricals). However, if a matrix is constructed with the new `cat_missing_method` parameter set to zero, then missing categorical values are allowed, and are treated as all-zero indicator rows in subsequent method calls. This new behavior is consistent with `pandas.get_dummies`' way of handling missing values.

Notes about performance/templating functions:

 - When a categorical matrix has missing values, the `*_complex` (formerly `*_drop_first`) suite of functions is used, so this feature has no performance impact in the most important case (no drop_first and no missing values).
 - Most of the changes on the C++/Cython side are very minor. The only bigger differences are in the `_sandwich_cat_dense*` and `_sandwich_cat_cat*` functions, which were not templated by `drop_first` before, but are now.
 - Benchmarks show that performance differences between the `*_fast` and `*_complex` functions is negligible in most cases. The only exception is `CategoricalMatrix.tocsr` (and subsequently `CategoricalMatrix.multiply`), where the `_*fast` function is indeed faster (by 25% and 10%, respectively, on a 10,000,000 x 5 categorical matrix). This might imply that having two versions of these functions (like the newly templated `Categorical._sandwich_cat_dense*` and `Categorical._sandwich_cat_cat`, but maybe even `Categorical.sandwich`, `Categorical.matvec` and `Categorical.transpose_matvec`) may add more complexity than the benefit it brings.

### DIFF
--- a/src/tabmat/categorical_matrix.py
+++ b/src/tabmat/categorical_matrix.py
@@ -617,7 +617,7 @@ class CategoricalMatrix(MatrixBase):
                 dtype=self.dtype,
                 column_name=self._colname,
                 column_name_format=self._colname_format,
-                cat_missing_method=self.missing_method,
+                cat_missing_method=self._missing_method,
             )
         else:
             # return a SparseMatrix if we subset columns
--- a/src/tabmat/categorical_matrix.py
+++ b/src/tabmat/categorical_matrix.py
@@ -298,7 +298,11 @@ class CategoricalMatrix(MatrixBase):
 
         Test: matrix/test_categorical_matrix::test_recover_orig
         """
-        return self.cat.categories[self.cat.codes]
+        orig = self.cat.categories[self.cat.codes].to_numpy()
+        if self.has_missing:
+            orig = orig.view(np.ma.MaskedArray)
+            orig.mask = self.cat.codes == -1
+        return orig
 
     def _matvec_setup(
         self,
--- a/src/tabmat/categorical_matrix.py
+++ b/src/tabmat/categorical_matrix.py
@@ -237,7 +237,7 @@ class CategoricalMatrix(MatrixBase):
         drop the first level of the dummy encoding. This allows a CategoricalMatrix
         to be used in an unregularized setting.
 
-    missing_method: str {'fail'|'ignore'}, default 'fail'
+    cat_missing_method: str {'fail'|'ignore'}, default 'fail'
         How to handle missing values. Either "fail" or "zero". If "fail", an error
         will be raised if there are missing values. If "zero", missing values will
         represent all-zero indicator columns.
--- a/src/tabmat/categorical_matrix.py
+++ b/src/tabmat/categorical_matrix.py
@@ -585,6 +585,7 @@ class CategoricalMatrix(MatrixBase):
                 dtype=self.dtype,
                 column_name=self._colname,
                 column_name_format=self._colname_format,
+                cat_missing_method=self.missing_method,
             )
         else:
             # return a SparseMatrix if we subset columns
--- a/src/tabmat/categorical_matrix.py
+++ b/src/tabmat/categorical_matrix.py
@@ -613,15 +613,17 @@ class CategoricalMatrix(MatrixBase):
 
         res = sandwich_cat_dense(
             self.indices,
-            self.shape[1] + self.drop_first,
+            self.shape[1],
             d,
             other,
             rows,
             R_cols,
             is_c_contiguous,
+            has_missing=self.has_missing,
+            drop_first=self.drop_first,
         )
 
-        res = _row_col_indexing(res[self.drop_first :], L_cols, None)
+        res = _row_col_indexing(res, L_cols, None)
         return res
 
     def _cross_categorical(
@@ -649,6 +651,8 @@ class CategoricalMatrix(MatrixBase):
             d.dtype,
             self.drop_first,
             other.drop_first,
+            self.has_missing,
+            other.has_missing,
         )
 
         res = _row_col_indexing(res, L_cols, R_cols)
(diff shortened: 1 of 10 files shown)


lbittarello:
What about adding the option to treat missing values as category in their own? :)

MartinStancsicsQC:
Yes, that's a great idea, and I believe it only takes a few extra lines in the `__init__` methods to implement. And I believe `CategoricalMatrix` would then have full feature parity with `pandas.get_dummies`. I'll add it this afternoon.

MartinStancsicsQC:
I've added the option to treat missings in categorical columns as their own category. What do you think, @lbittarello?

Couple of notes:
 - This is implemented as a pre-processing step instead of complicating things on the Cython/C++ side. The only drawback is that we need to make a copy of categorical columns with missing values, but I don't think that is a big issue.
 - In contrast to `pandas.get_dummies`, a new category is only added if the column actually contains missing values. I think this is more reasonable for our purposes, as all-zero columns make no sense when estimating a model.

src/tabmat/categorical_matrix.py
```diff
@@ -241,6 +241,11 @@ class CategoricalMatrix(MatrixBase):
         drop the first level of the dummy encoding. This allows a CategoricalMatrix
         to be used in an unregularized setting.
 
+    cat_missing_method: str {'fail'|'zero'|'convert'}, default 'fail'
+        - if 'fail', raise an error if there are missing values
```
MatthiasSchmidtblaicherQC, line 245: missing "."

src/tabmat/categorical_matrix.py
```diff
@@ -250,15 +255,43 @@ def __init__(
...
+            if self._missing_method == "fail":
+                raise ValueError(
+                    "Categorical data can't have missing values "
+                    "if cat_missing_method='fail'."
+                )
+
+            elif self._missing_method == "convert":
+                if self._missing_category not in self.cat.categories:
+                    self.cat = self.cat.add_categories([self._missing_category])
+                else:
```
MatthiasSchmidtblaicherQC, line 283: This is not really consistent with `pd.get_dummies`. If I have a category`(MISSING)` in a pandas series, then `pd.get_dummies` will not encode 1's for the NA's in the same columns as for category `(MISSING)`. 
```
x = pd.Series(["x", "y", "((MISSING)", pd.NA])
pd.get_dummies(x, dummy_na=True)
pd.get_dummies(x)
```
I think I would rather raise if `(MISSING)` is already a category...

MatthiasSchmidtblaicherQC (commented):
We should clarify under which circumstances we want to have the same behaviour as `pd.get_dummies` (except for not expanding the dummy matrix), and where we want to deviate. Otherwise, looks great!

src/tabmat/categorical_matrix.py
MartinStancsicsQC: Good catch, thank you. I agree, raising would be a better/safer default.

MartinStancsicsQC: @MatthiasSchmidtblaicherQC, what do you think of the current solution? (raise if missing category exists, but also parametrize `_missing_category`)

MartinStancsicsQC:
@MarcAntoineSchmidtQC, let me know if we should merge it into main or aim for the next major release/refactor

src/tabmat/categorical_matrix.py
```diff
@@ -250,15 +260,41 @@ def __init__(
...
         drop_first: bool = False,
         dtype: np.dtype = np.float64,
+        cat_missing_method: str = "fail",
+        cat_missing_name: str = "(MISSING)",
     ):
-        if pd.isnull(cat_vec).any():
-            raise ValueError("Categorical data can't have missing values.")
+        if cat_missing_method not in ["fail", "zero", "convert"]:
+            raise ValueError(
+                "cat_missing_method must be 'fail' 'zero' od 'convert', "
```
MatthiasSchmidtblaicherQC, line 268:
```suggestion
                "cat_missing_method must be one 'fail' 'zero' or 'convert', "
```

src/tabmat/categorical_matrix.py
```diff
@@ -250,15 +260,41 @@ def __init__(
...
+                )
+
+            elif self._missing_method == "convert":
+                self.cat = self.cat.add_categories([self._missing_category])
+
+                self.cat[pd.isnull(self.cat)] = self._missing_category
+                self._has_missing = False
+
+            else:
+                self._has_missing = True
```
MatthiasSchmidtblaicherQC, line 293: has_missing or has_missings?

src/tabmat/categorical_matrix.py
MatthiasSchmidtblaicherQC: I like the solution. What error message does `self.cat = self.cat.add_categories([self._missing_category])` give if it fails because the missing category is already part of the categories? If it is unintuitive, we might want to raise an error with customised message in this case.

MartinStancsicsQC: It raises `ValueError: new categories must not include old categories: {'(MISSING)'}`, which I thought was okay, but I can of course replace it with a more tabmat-specific one.

src/tabmat/categorical_matrix.py
MartinStancsicsQC: Hm, has_missings indeed sounds a bit more natural.

MarcAntoineSchmidtQC (approved):
Amazing. The tests seem to cover a lot. I don't have anything to add here.

MarcAntoineSchmidtQC:
We can merge to main right now. My understanding is that it's fully backward compatible.

PR #286, opened 2023-08-09
Tabmat v4 alpha

MartinStancsicsQC:
Checklist

The core changes for the Tabmat 4.0 release. The main goal is making the API of `MatrixBase subclasses more consistent. All other PRs planned for the 4.0 release are based on this (`tabmat-v4`) branch.

## Major changes

- `DenseMatrix` and `SparseMatrix` are not subclasses of numpy ans scipy arrays, anymore. It means that they do not inherit their (sometimes conflicting) behavior, and only expose a minimal interface (i.e. mostly the methods that `SplitMatrix` exposes).
- `DenseMatrix` and `SparseMatrix` now contain their data in an `_array` attribute, containing a `numpy.ndarray` or a `scipy.csc_matrix`, respectively. The underlying data structures can be accessed using the `.unpack()` method.
- `MatrixBase` objects are always 2-dimensional (this was not the case with `DenseMatrix` before. One-dimensional inputs to their constructors are interpreted as column matrices.


MartinStancsicsQC:
It seems that our CI is set up to [publish to PyPI on any GH release](https://github.com/Quantco/tabmat/blob/fbf7755aa48ada1d72dec5d61b7fd9d9f23fdcc6/.github/workflows/build_wheels_release.yml#L3-L6) (including pre-releases), so creating a release with a corresponding alpha/beta/rc tag should be enough to get the v4 branch to PyPI.

src/tabmat/split_matrix.py
```diff
@@ -448,3 +485,60 @@ def __repr__(self):
...
+    def get_names(
+        self,
+        type: str = "column",
+        missing_prefix: Optional[str] = None,
+        indices: Optional[List[int]] = None,
+    ) -> List[Optional[str]]:
+        """Get column names.
+
+        For columns that do not have a name, a default name is created using the
+        followig pattern: ``"{missing_prefix}{start_index + i}"`` where ``i`` is
```
MatthiasSchmidtblaicherQC, line 498:
```suggestion
        following pattern: ``"{missing_prefix}{start_index + i}"`` where ``i`` is
```

src/tabmat/split_matrix.py
```diff
@@ -448,3 +485,60 @@ def __repr__(self):
...
+        the index of the column.
+
+        Parameters
+        ----------
+        type: str {'column'|'term'}
+            Whether to get column names or term names. The main difference is that
+            a categorical submatrix is counted as a single term, whereas it is
```
MatthiasSchmidtblaicherQC, line 505:
```suggestion
            a categorical submatrix has a single term but may have multiple columns.
```

src/tabmat/split_matrix.py
```diff
@@ -448,3 +485,60 @@ def __repr__(self):
...
+            counted as multiple columns. Furthermore, matrices created from formulas
+            have a difference between a column and term (c.f. ``formulaic`` docs).
```
MatthiasSchmidtblaicherQC, line 507:
```suggestion
            distinguish between a column and term (c.f. ``formulaic`` docs).
```

stanmart (approved):
Looks good!

setup.py
```diff
@@ -54,7 +54,7 @@
 print(f"Debug Build: {debug_build}")
 
 if sys.platform == "win32":
-    allocator_libs = []
+    allocator_libs = []  # type: ignore
```
MarcAntoineSchmidtQC, line 57: Can we remove this? Doesn't seem to be a problem currently. Maybe an outdated pre-commit hook?

src/tabmat/__init__.py
```diff
@@ -21,5 +21,8 @@
     "SplitMatrix",
     "CategoricalMatrix",
     "from_csc",
+    "from_formula",
     "from_pandas",
+    "as_tabmat",
+    "hstack",
```
MarcAntoineSchmidtQC, line 27: are we sure "hstack" is something we want to expose at the core of tabmat? I think it would be easier for people to import it using `from tabmat.split_matrix import hstack`

CHANGELOG.rst
```diff
@@ -10,6 +10,13 @@ Changelog
 Unreleased
 ----------
 
+**New features:**
+
+- Added column name and term name metadata to ``MatrixBase`` objects. These are automatically populated when initializing a ``MatrixBase`` from a ``pandas.DataFrame``. In addition, they can be accessed and modified via the ``column_names`` and ``term_names`` properties.
+- Added a formula interface for creating tabmat matrices from pandas data frames. See :func:`tabmat.from_formula` for details.
+- Added support for missing values in ``CategoricalMatrix`` by either creating a separate category for them or treating them as all-zero rows.
+- Added support for handling missing categorical values in pandas data frames.
+
```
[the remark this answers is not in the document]
MarcAntoineSchmidtQC: Also, we should describe if there are any steps needed to migrate from V3 to V4

src/tabmat/__init__.py
stanmart: It's a bit misleading that `hstack` lives in the `split_matrix` module, because it works on all `MatrixBase` subclasses, and returns the expected type for all combinations. I.e., its signature is
```
hstack(tup: Sequence[Union[MatrixBase, np.ndarray, sps.spmatrix]]) -> MatrixBase
```

CHANGELOG.rst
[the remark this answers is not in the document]
stanmart: Good point. IIRC the glum-v3 core branch was just for making glum compatible with tabmat v4, without adding any new features. So looking at [this diff](https://github.com/Quantco/glum/compare/main...glum-v3-core) might give a good idea of the backward incompatible changes.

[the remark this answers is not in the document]
MatthiasSchmidtblaicherQC: What about something like [this](https://github.com/Quantco/tabmat/pull/286/commits/4d621d6b10f0f82674c5c43567b090e1dd96071f).

src/tabmat/ext/cat_split_helpers-tmpl.cpp
```diff
@@ -15,16 +15,16 @@ void _transpose_matvec_${type}(
 ) {
     int num_threads = omp_get_max_threads();
     std::vector<F> all_res(num_threads * res_size, 0.0);
-    #pragma omp parallel shared(all_res) 
+    #pragma omp parallel shared(all_res)
```
MatthiasSchmidtblaicherQC, line 18: I incorporated the changes from https://github.com/Quantco/tabmat/pull/348 here. @adityagoel4512 @stanmart would you mind having a quick look at the changes?

adityagoel4512: Looks fine from my perspective. The nightly failures look unrelated to the change.

MatthiasSchmidtblaicherQC: Thanks for checking. Yes, I think the nightly failures are related to the install failure for 3.9 in https://github.com/Quantco/tabmat/issues/350 - a separate issue.

src/tabmat/ext/cat_split_helpers-tmpl.cpp
```diff
@@ -1,26 +1,29 @@
...
 	int tid = omp_get_thread_num();
-	F* res_slice = &all_res[tid * res_size]; 
+	F* res_slice = &all_res[tid * res_size];
 	#pragma omp for
         for (Py_ssize_t i = 0; i < n_rows; i++) {
-            % if dropfirst == 'all_rows_drop_first':
-                Py_ssize_t col_idx = indices[i] - 1;
-                if (col_idx != -1) {
+            % if type == 'all_rows_complex':
+                Py_ssize_t col_idx = indices[i] - drop_first;
```
adityagoel4512, line 25:
```suggestion
                Py_ssize_t col_idx = indices[i] - static_cast<Int>(drop_first);
```

I don't think it really makes a difference right now, but it might be better to be explicit in case this is refactored in the future.

stanmart:
🎉

PR #290, opened 2023-08-14
Explicitly link to OpenMP

xhochy:
Fixes #220 

Checklist
* [ ] Added a `CHANGELOG.rst` entry

PR #293, opened 2023-08-16
Major CI cleanup

pavelzw:
Similar structure as in https://github.com/Quantco/glum/pull/674


pavelzw:
Just saw #291 :)

.github/workflows/build-wheels.yml
```diff
@@ -15,16 +16,19 @@ jobs:
...
         uses: docker/setup-qemu-action@v2
         with:
           platforms: all
+
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.14.1
         env:
           CIBW_ARCHS_LINUX: auto aarch64
           CIBW_ARCHS_MACOS: x86_64 arm64
+          CIBW_TEST_SKIP: '*-macosx_arm64'
```
pavelzw, line 31: This is from #290 where it was not introduced in all places.

.github/workflows/ci.yml
```diff
@@ -2,64 +2,63 @@ name: CI
...
-          - linux_64_python3.7_default.____cpython
-          - linux_aarch64_python3.10_default.____cpython
-          - linux_ppc64le_python3.10_default.____cpython
+        include: # TODO: 3.7? what about 3.11? is every case covered?
+          - { os: ubuntu-latest, python-version: '3.8' } # before we had 3.7, 3.8, 3.9, 3.10 in ubuntu-latest
+          - { os: ubuntu-latest, python-version: '3.9' }
+          - { os: ubuntu-latest, python-version: '3.10' }
+          - { os: ubuntu-latest, python-version: '3.11' }
+          - { os: macos-latest, python-version: '3.10' } # before we had 3.7 in macos-latest
+          - { os: windows-latest, python-version: '3.10' } # before we had 3.7 in windows-latest
```
pavelzw, line 37: Which python versions do we want to test here?

.github/workflows/conda-build.yml
```diff
@@ -0,0 +1,61 @@
...
+          - { conda_build_yml: linux_64_python3.7_default.____cpython,  os: ubuntu-latest,  conda-build-args: '' }
+          - { conda_build_yml: linux_64_python3.8_default.____cpython,  os: ubuntu-latest,  conda-build-args: '' }
+          - { conda_build_yml: linux_64_python3.9_default.____cpython,  os: ubuntu-latest,  conda-build-args: '' }
+          - { conda_build_yml: linux_64_python3.10_default.____cpython, os: ubuntu-latest,  conda-build-args: '' }
+          - { conda_build_yml: osx_64_python3.7_default.____cpython,    os: macos-latest,   conda-build-args: '' }
+          - { conda_build_yml: osx_64_python3.8_default.____cpython,    os: macos-latest,   conda-build-args: '' }
+          - { conda_build_yml: osx_arm64_python3.9.____cpython,         os: macos-latest,   conda-build-args: ' --no-test' }
+          - { conda_build_yml: osx_arm64_python3.10.____cpython,        os: macos-latest,   conda-build-args: ' --no-test' }
+          - { conda_build_yml: win_64_python3.9.____cpython,            os: windows-latest, conda-build-args: '' }
+          - { conda_build_yml: win_64_python3.10.____cpython,           os: windows-latest, conda-build-args: '' }
```
pavelzw, line 27: This is copied from the various conda build workflows. Not all variants are tested in CI, like `linux_aarch64_python3.10_default.____cpython`. Should we also test aarch64 and ppc64le in CI?

pavelzw:
Should we add this part to the tests?

```yml
      - name: Run doctest
        shell: bash -el {0}
        # Check that the readme example will work by running via doctest.
        # We run outside the repo to make the test a bit more similar to
        # a user running after installing with conda.
        run: |
          mkdir ../temp
          cp README.md ../temp
          cd ../temp
          python -m doctest -v README.md
```

README.md
```diff
@@ -1,6 +1,10 @@
 # Efficient matrix representations for working with tabular data
 
-![CI](https://github.com/Quantco/tabmat/workflows/CI/badge.svg)
+[![CI](https://github.com/Quantco/tabmat/workflows/CI/badge.svg)](https://github.com/Quantco/tabmat/actions)
+[![Docs](https://readthedocs.org/projects/pip/badge/?version=latest&style=flat)](https://tabmat.readthedocs.io/)
+[![Conda-forge](https://img.shields.io/conda/vn/conda-forge/tabmat?logoColor=white&logo=conda-forge)](https://anaconda.org/conda-forge/tabmat)
+[![PypiVersion](https://img.shields.io/pypi/v/tabmat.svg?logo=pypi&logoColor=white)](https://pypi.org/project/tabmat)
+[![PythonVersion](https://img.shields.io/pypi/pyversions/tabmat?logoColor=white&logo=python)](https://pypi.org/project/tabmat)
```
MarcAntoineSchmidtQC, line 7: Thanks!

.github/workflows/ci.yml
MarcAntoineSchmidtQC: We still want to support 3.7, so we should test it.

MarcAntoineSchmidtQC:
> ```yaml
>  the readme example will work by running via doctest.
>         # We run outside the repo to make the test a bit more similar to
>         # a user run
> ```

I don't see a reason why not. It's also in glum, right?

.github/workflows/conda-build.yml
pavelzw: I'll try it out.

pavelzw: ~6min seems fine. Quicker than macOS and Windows 😅

MarcAntoineSchmidtQC: Yes, please add them.

MarcAntoineSchmidtQC (approved):
Final change I made was to revert back to only building wheel for aarch64 on release.

In a future PR I would like to fully support 3.11.

PR #312, opened 2023-10-11
Add 3.12 to metadata remove 3.7 and 3.8

jtilly:
In line with #307


xhochy:
Personally, I would prefer to have #311 merged.

PR #294, opened 2023-08-16
Use micromamba to create conda environments in Wheel build

xhochy:
Checklist
* [ ] Added a `CHANGELOG.rst` entry

PR #315, opened 2023-10-14
Revert to conda wheel build

MarcAntoineSchmidtQC:
This PR needs some background information. 
The wheels for macos-arm64 had been broken prior to the 3.1.11 release. #290 fixed the problem, but #294 seemed to have broken it again. The problem can be seen from the delocate (wheel repair) part of cibuildwheel. We need to link three dynamic libraries (libc++, jemalloc, and libomp). With #294, introducing micromamba in the build of the wheel, jemalloc has not been packaged with the wheel anymore.

To be honest, I'm not exactly sure what makes micromamba different from conda. 

Here's where we can find the difference: https://github.com/Quantco/tabmat/actions/runs/6514998408/job/17696892050#step:4:684

You can see that we were missing one library before (this is from the last 3.1.11 release):
https://github.com/Quantco/tabmat/actions/runs/6511122493/job/17686217662#step:6:523

@xhochy, if you know why micromamba is acting this way and how we can fix it, I'm happy to close this PR and fix the problem differently.

Checklist

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -21,9 +21,6 @@ jobs:
         uses: docker/setup-qemu-action@v3
         with:
           platforms: all
-
-      - name: Set up micromamba
-        uses: mamba-org/setup-micromamba@2b72821d5ad7f6da3c003a3684ce341bf187b46f
       
       - name: Build wheels
         if: github.event_name != 'release'
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -1,6 +1,7 @@
 name: Build and upload to PyPI
 
 on:
+  push:
   pull_request:
   release:
     types:
--- a/build_tools/prepare_macos_wheel.sh
+++ b/build_tools/prepare_macos_wheel.sh
@@ -8,4 +8,4 @@ else
     CONDA_CHANNEL="conda-forge/osx-64"
 fi
 
-/Users/runner/micromamba-bin/micromamba create -y -p $CONDA/envs/build -c $CONDA_CHANNEL jemalloc-local xsimd llvm-openmp
+conda create -y -n build -c $CONDA_CHANNEL jemalloc-local xsimd llvm-openmp


jtilly (approved):

jtilly:
@jonashaag do you happen to know why conda and micromamba do different things here?

ivergara:
Ping @pavelzw

jonashaag:
The only idea I have is that `setup-micromamba` had some side effects on the environment. Typically that's changing the shell config. Let's see if that's the case here https://github.com/Quantco/tabmat/tree/jonashaag-patch-1

jtilly:
> The only idea I have is that `setup-micromamba` had some side effects on the environment. Typically that's changing the shell config. Let's see if that's the case here https://github.com/Quantco/tabmat/tree/jonashaag-patch-1

Ah, interesting hypothesis. That doesn't seem to be it though. When I look at the [output from your run](https://github.com/Quantco/tabmat/actions/runs/6518571635/job/17704132197#step:5:522), I also don't see `libjemalloclocal` for the arm64 runs (only for the x86 runs).

jonashaag:
Let's see what Mamba does https://github.com/Quantco/tabmat/tree/jonashaag-patch-2

jtilly:
Is it possible that micromamba and conda diverge in how they treat `-c "conda-forge/osx-arm64` in our [prepare build script](https://github.com/Quantco/tabmat/blob/e65e4e03b15811c2c11b49208674d1df9868b563/build_tools/prepare_macos_wheel.sh#L5-L9)?

The two GitHub action runs that @MarcAntoineSchmidtQC is linking to above install slightly different versions of `jemalloc`:

[5.3.0  hf0c8a7f_0](https://github.com/Quantco/tabmat/actions/runs/6511122493/job/17686217662#step:6:457) (the osx-64 build ❌ ) vs [5.3.0 hb7217d7_0](https://github.com/Quantco/tabmat/actions/runs/6514998408/job/17696892050#step:4:542) (the osx-arm64 build ✅ )

(see https://anaconda.org/conda-forge/libjemalloc-local/files?version=5.3.0)

Setting the subdir as environment variable ([commit](https://github.com/Quantco/tabmat/commit/1b696a2dbcea0bf87c306d362d70a515f3e12416), [GH actions](https://github.com/Quantco/tabmat/actions/runs/6519300678/job/17705631090#step:5:528)) appears to work.

jonashaag:
Good catch! Indeed, this is a well-known issue https://github.com/mamba-org/mamba/issues/2041. @ThomasBlauthQC once attempted to fix this but hasn't gotten around to actually fixing it, see https://github.com/mamba-org/mamba/issues/2789

MarcAntoineSchmidtQC:
I'm happy that we figured this one out! Thanks everyone.

I looked at the runtime and, perhaps unsurprisingly, micromamba is faster. Since we are using micromamba pretty much everywhere else in our projects, I lean toward using this solution.

@jtilly, any thoughts? If you agree, I'll open up a PR based on your `micromamba` branch.

jtilly:
> any thoughts? If you agree, I'll open up a PR based on your micromamba branch.

Yes, that sounds good!

MarcAntoineSchmidtQC:
Replaced by #317

PR #326, opened 2023-10-30
Use ruff as our formatter/linter

MarcAntoineSchmidtQC:
Following the widespread adoption of ruff at QuantCo, this PR changes the linter AND formatter system to use [ruff](https://docs.astral.sh/ruff/). This replaces black, flake8, isort, and pyupgrade.

Most changes to the code are from line-too-long errors (strings).

Checklist


.pre-commit-config.yaml
```diff
@@ -1,46 +1,18 @@
...
+    rev: 0.1.3
+    hooks:
+      - id: ruff-conda
+      - id: ruff-format-conda
+  - repo: https://github.com/Quantco/pre-commit-mirrors-mypy
+    rev: "1.5.1"
+    hooks:
+      - id: mypy-conda
+        additional_dependencies:
+          - python=3.9
```
jtilly, line 12:
```suggestion
```
Let's use this opportunity to get rid of this here and instead specify
```
[tool.mypy]
python_version = '3.9'
```
in `pyproject.toml`. The Python version that's used to run `mypy` doesn't have to be the same as the Python version that we're targeting.

jtilly (approved):

PR #335, opened 2023-12-11
Fix xsimd issue with ppc64le

MarcAntoineSchmidtQC:
xsimd does play along nicely with the ppc64le architecture. This has already been flagged, so we will probably be able to use xsimd>=12.2 once available.

Checklist


MarcAntoineSchmidtQC:
Also fixes the macos wheel issue. Should be all green.

jtilly (approved):

PR #348, opened 2024-02-23
CategoricalMatrix A.Tb reproducibility.

adityagoel4512:
The bit reproducibility of `CategoricalMatrix`'s `transpose_matvec` can't be guaranteed since different threads can increment same entry in the output vector - since fp addition is not associative, the output will not necessarily be the same every time. We can get this level of reproducibility if each slice has one thread associated to it which is then aggregated deterministically (parallelising over the number of categories when aggregating into the output).

This is not entirely for free (result below on my Macbook with spawns 12 threads), so I'll leave this to regular users to determine if this makes more sense behind a flag or if it is an acceptable trade off:
```bash
python src/tabmat/benchmark/main.py --one_cat 1000000 1000000 --n_iterations 50
After:
1  transpose-matvec  tabmat  0.004588  dense_smallcat
Before (main):
1  transpose-matvec  tabmat  0.003998  dense_smallcat
```
Checklist


src/tabmat/ext/cat_split_helpers-tmpl.cpp
```diff
@@ -10,24 +10,29 @@ void _transpose_matvec_${dropfirst}(
     F* res,
     Int res_size
 ) {
-    #pragma omp parallel
+    int num_threads = omp_get_max_threads();
+    std::vector<F> all_res(num_threads * res_size, 0.0);
```
MarcAntoineSchmidtQC, line 14: Am I understanding correctly that we are using ~N times the memory compared to before (N being the number of threads)?
For transpose matvec this can be a problem because the result is as large as the input. Therefore, if the user has a very large categorical matrix as the main part of X, it will require N times more memory than X.

adityagoel4512: I'm not the most familiar with OpenMP but was a new vector `restemp` not allocated per thread before (see https://github.com/Quantco/tabmat/blob/main/src/tabmat/ext/cat_split_helpers-tmpl.cpp#L15)?

MarcAntoineSchmidtQC: You are right. Then let's merge this. Thanks for the contribution!

PR #356, opened 2024-04-11
Set fill value of sparse columns in line with ``SparseDtype.subtype``

MatthiasSchmidtblaicherQC:
See https://github.com/Quantco/glum/issues/784.

Checklist

### DIFF
--- a/src/tabmat/constructor.py
+++ b/src/tabmat/constructor.py
@@ -108,10 +108,7 @@ def from_pandas(
             # check if we want to store as sparse
             if (coldata != 0).mean() <= sparse_threshold:
                 if not isinstance(coldata.dtype, pd.SparseDtype):
-                    if is_bool_dtype(coldata):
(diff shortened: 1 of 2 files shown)


lbittarello (approved):

src/tabmat/constructor.py
```diff
@@ -108,7 +108,11 @@ def from_pandas(
             # check if we want to store as sparse
             if (coldata != 0).mean() <= sparse_threshold:
                 if not isinstance(coldata.dtype, pd.SparseDtype):
-                    sparse_dtype = pd.SparseDtype(coldata.dtype, fill_value=0)
+                    if is_bool_dtype(coldata):
+                        fill_value = False
+                    else:
+                        fill_value = 0
```
lbittarello, line 114:
```suggestion
                    fill_value = False if is_bool_dtype(coldata) else 0
```

PR #366, opened 2024-05-27
Fix osx wheel

MarcAntoineSchmidtQC:
Checklist

### DIFF
--- a/build_tools/prepare_macos_wheel.sh
+++ b/build_tools/prepare_macos_wheel.sh
@@ -8,4 +8,4 @@ else
     export CONDA_SUBDIR="osx-64"
 fi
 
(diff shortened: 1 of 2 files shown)


MarcAntoineSchmidtQC:
@xhochy, I saw that you were recently involved with some PR on the llvm-openmp feedstock. Do you have any idea why the 18.1.6 release seems to have bumped the minimum target version to 10.13?

MarcAntoineSchmidtQC:
Recap: we use conda-forge to install some build dependencies (including llvm-openmp). Conda-forge recently bumped the deployment target to 10.13, and the latest version of llvm-openmp was built using macos 10.13.

I think the best solution is to use 10.13 as the deployment version. This will avoid more issues with other dependencies.

jtilly (approved):

PR #370, opened 2024-06-17
`from_polars`

lbittarello:
This PR adds basic support for Polars.

On the surface level, it adds a `from_polars` function, which mirrors `from_pandas` (except in that it doesn't allow for object columns, which are painful to work with in Polars). Because accessor methods differ between Pandas and Polars, it also restructures the `CategoricalMatrix` so it stores categories instead of a categorical array (which should also lower its memory footprint).

This PR also streamlines `from_pandas` and collects tests into a dedicated file.

I haven't extended the formula interface, which will be left to a future PR.


MarcAntoineSchmidtQC:
Trying to fix the build error with numpy 2.0. In the meantime, you are missing the dependency in `setup.py` (line 160).

src/tabmat/constructor.py
```diff
@@ -133,53 +133,205 @@ def from_pandas(
...
+                mxcolidx += 1
+            else:
+                dense_columns.append(coldata.name)
+                dense_indices.append(mxcolidx)
+                mxcolidx += 1
+
+        else:
+            ignored_cols.append(coldata.name)
+
+    if len(ignored_cols) > 0:
```
cbourjau, line 311: This later part looks very similar to the Pandas version. Might we be able to ducktype this with a shared helper function?

cbourjau (commented):
Interesting to see some Pandas/Polars compatibility in action! I left a few comments as ideas to possibly decrease code duplication, if you haven't yet considered them.

src/tabmat/categorical_matrix.py
```diff
@@ -302,38 +321,50 @@ def __init__(
             self._has_missings = False
 
         self.drop_first = drop_first
-        self.shape = (len(self.cat), len(self.cat.categories) - int(drop_first))
-        self.indices = self.cat.codes.astype(np.int32)
-        self.x_csc: Optional[tuple[Optional[np.ndarray], np.ndarray, np.ndarray]] = None
+        self.indices = indices.astype(np.int32)
```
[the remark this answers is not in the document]
lbittarello: `indices` is an existing attribute, so we need to preserve it for backward compatibility.

src/tabmat/constructor.py
```diff
@@ -133,53 +133,205 @@ def from_pandas(
                     mxcolidx += cat.shape[1]
                 elif cat_position == "end":
                     indices.append(np.arange(cat.shape[1]))
-        # All other numerical dtypes (needs to be after pd.SparseDtype)
         elif is_numeric_dtype(coldata):
-            # check if we want to store as sparse
             if (coldata != 0).mean() <= sparse_threshold:
-                if not isinstance(coldata.dtype, pd.SparseDtype):
```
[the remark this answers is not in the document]
lbittarello: Exactly 🚀

src/tabmat/constructor.py
```diff
@@ -133,53 +133,205 @@ def from_pandas(
...
+
+    if len(matrices) > 1:
+        return SplitMatrix(matrices, indices)
+    elif len(matrices) == 0:
+        raise ValueError("DataFrame contained no valid column")
+    else:
+        return matrices[0]
+
+
+def from_polars(
```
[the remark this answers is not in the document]
lbittarello: I have the impression that we want to keep dependencies to a minimum, especially dependencies on experimental libraries, so I don't know that we want to call Narwhals or require users to call Narwhals.

src/tabmat/constructor.py
lbittarello: Happy to add a helper function, but is it worth it? It'd require lots of inputs and would either modify them in place or return lots of outputs, both of which make the code less transparent.

cbourjau: Would it be so many inputs and outputs? Can't the helper function just return the single matrix at the end? I'd say it is fine for a private helper function to modify the inputs if it significantly cuts down on the duplicated code.

src/tabmat/categorical_matrix.py
```diff
@@ -302,38 +351,56 @@ def __init__(
...
+        """Return a series with same data as what was initially fed to __init__.
+
+        This property is available for backward compatibility.
+        """
+        warnings.warn(
+            "This property will be removed in the next major release.",
+            category=DeprecationWarning,
+        )
+
+        if _is_polars(self._input_dtype):
```
cbourjau, line 380: Am I mistaken that `_input_dtype` may neither be "polars" nor "pandas" if the initial array given to `__int__` was a plain NumPy array?

src/tabmat/categorical_matrix.py
```diff
@@ -255,7 +283,7 @@ class CategoricalMatrix(MatrixBase):
 
     def __init__(
         self,
-        cat_vec: Union[list, np.ndarray, pd.Categorical],
+        cat_vec,
```
cbourjau, line 286: mypy might actually be useful here. It is quite good with union types and narrowing them in if-statements these days.

cbourjau (commented):
I suppose we are breaking new grounds by supporting polars, pandas and NumPy in here. I think/hope we might find a slightly easier-to-maintain approach still. Here are some thoughts:

I think some of the difficulties originate from having the mapping of different backend libraries intertwined with the business logic of this library. Things might get easier if the mapping is more cleanly separated behind wrapper classes. The `CategoricalMatrix` takes a `cat_vec` as an argument. Right now, the `__init__` method maps all possible input types to the disassembled representation of a tuple of NumPy arrays+dtype (namely `indices`, `categories`, and `_input_dtype`). The `cat` property then reassembles these with a similar mapping. Instead, one may consider wrapping the `cat_vec`  in a class that exposes exactly the data that we might want to derive from `cat_vec` as properties/functions. These functions would encapsulate the necessary mapping logic. Something like:

```python
class CatVec:
    _wrapped: np.ndarray | pd.Series | pl.Series
    
    @property
    def categories(self) -> np.ndarray:
        # get categories from `self._wrapped` on the fly
        ...
        
    @property
    def codes(self) -> np.ndarray: ...
    
    @property
    def shape(self) -> tuple[int, ...]: ...
    
    def contains_missing(self) -> bool: ...
```

The `CategoricalMatrix.index`, `CategoricalMatrix.shape`, etc members could be replaced by properties that simply call through to the (privately) stored `CatVec` instance. Might this work better?

lbittarello:
I appreciate that the current code is messy and wrapper classes would help streamline it, but it'll look better once we are past the transition phase.

We currently have a `cat` attribute that returns a `pandas.Categorical` type, regardless of the input type. This PR transformed it into a property, which returns a `polars.Series` if the input is one and a `pandas.Categorical` otherwise. We use the `_input_dtype` attribute to keep track of the input type.

At some point, we will do away with the `cat` property. At that point, we'll also be able to remove the `_Categorical` helper class or the `_input_dtype` attribute. We'll only need to extract the categories and indices from the input vector, which is straightforward.

If your concern is maintenance, it's all temporary, so it might not warrant a lot of infrastructure.

src/tabmat/categorical_matrix.py
lbittarello: `mypy` doesn't play nice with optional imports. I tried conditioning imports on type hinting to no avail.

src/tabmat/categorical_matrix.py
```diff
@@ -264,35 +292,56 @@ def __init__(
...
+            self.categories = cat_vec.cat.categories.to_numpy()
+            indices = cat_vec.cat.codes.to_numpy()
+        elif _is_polars(cat_vec):
+            if not _is_polars(cat_vec.dtype):
+                cat_vec = cat_vec.cast(pl.Categorical)
+            self.categories = cat_vec.cat.get_categories().to_numpy()
+            indices = cat_vec.to_physical().fill_null(-1).to_numpy()
         else:
-            self.cat = pd.Categorical(cat_vec)
+            indices, self.categories = pd.factorize(cat_vec, sort=True)
```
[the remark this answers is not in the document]
lbittarello: That's in line with current behaviour. The idea is for users to have either `pandas` or `polars`, not neither.

src/tabmat/categorical_matrix.py
lbittarello: That's in line with current behaviour and so backwards compatible. If the input is a NumPy array, we return a `pandas.Categorical`.

src/tabmat/categorical_matrix.py
cbourjau: Using [`TYPE_CHECKING`](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) didn't work? I think it is reasonable to assume both Pandas and Polars are installed in the dev environment, isn't it?

lbittarello: > Using [TYPE_CHECKING](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) didn't work?

No. :\ I may have done it wrong though.

cbourjau:
> I appreciate that the current code is messy and wrapper classes would help streamline it, but it'll look better once we are past the transition phase.

As you prefer

jtilly (approved):

src/tabmat/constructor.py
```diff
@@ -133,64 +131,177 @@ def from_pandas(
...
-        matrices.append(
-            DenseMatrix(
-                df.iloc[:, dense_dfidx].astype(dtype),
-                column_names=df.columns[dense_dfidx],
-                term_names=df.columns[dense_dfidx],
-            )
-        )
-        indices.append(dense_mxidx)
+    if dense_columns:
+        matrices.append(_dense_matrix(df, dense_columns, dtype))
```
MarcAntoineSchmidtQC, line 152: The [problem in glum](https://github.com/Quantco/glum/actions/runs/9772131161/job/26980639512) comes from here. It should be df[dense_columns]. However, using column names doesn't work here because of potential duplicates. This is why we had a logic to use the column index.

PR #371, opened 2024-06-17
Fix numpy 2.0 build error

MarcAntoineSchmidtQC:
Couple of changes from this PR:

- explicit numpy array import. This makes the error message more clear.
- replace "oldest-supported-numpy" with simply "numpy". We should probably make sure that this does not make it incompatible with older version of numpy.
- switch to macos-12 to built the macos wheel.
- add long long to numeric fused type. This probably has something to do with this new numpy feature from the 2.0 release: "The default integer type on Windows is now int64 rather than int32, matching the behavior on other platforms".

Checklist

Still to do:
* [x] Verify that the new build works with older versions of numpy

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ requires = [
   'setuptools-scm', 
   'wheel',
   'mako',
-  'oldest-supported-numpy',
+  'numpy',
   'Cython != 3.0.4',
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ requires = [
   'setuptools-scm', 
   'wheel',
   'mako',
-  'numpy',
+  'numpy>=1.25',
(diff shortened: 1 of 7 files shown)


MarcAntoineSchmidtQC:
The MacOS job is getting automatically canceled. Not sure what is going on.

MarcAntoineSchmidtQC:
well... macos-11 has been deprecated and removed...

pyproject.toml
```diff
@@ -4,7 +4,7 @@ requires = [
   'setuptools-scm', 
   'wheel',
   'mako',
-  'oldest-supported-numpy',
+  'numpy',
```
xhochy, line 7:
```suggestion
  'numpy>=1.25',
```

xhochy (changes requested):

xhochy (approved):

PR #376, opened 2024-07-03
Use column index in constructor

MarcAntoineSchmidtQC:
This fixes the failing tests in glum.

Checklist

### DIFF
--- a/src/tabmat/constructor.py
+++ b/src/tabmat/constructor.py
@@ -18,6 +18,190 @@ from .matrix_base import MatrixBase
 from .sparse_matrix import SparseMatrix
 from .split_matrix import SplitMatrix
 
+try:
+    import polars as pl
+except ImportError:
+    pl = None
+try:
+    import pandas as pd
+except ImportError:
+    pd = None
+
+
+def _is_numeric(series, engine: str):
+    if engine == "pandas":
+        return pd.api.types.is_numeric_dtype(series)
+    elif engine == "polars":
+        return series.dtype.is_numeric()
+    else:
+        raise ValueError(f"Unknown engine: {engine}")
+
+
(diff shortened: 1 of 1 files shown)


MarcAntoineSchmidtQC:
Still need to apply the same changes to polars

src/tabmat/constructor.py
```diff
@@ -148,13 +171,25 @@ def from_pandas(
         warnings.warn(
             f"Columns {ignored_cols} were ignored. Make sure they have a valid dtype."
         )
-    if dense_columns:
-        matrices.append(_dense_matrix(df, dense_columns, dtype))
-        indices.append(dense_indices)
+    if dense_dfidx:
+        matrices.append(
+            _dense_matrix(
```
MarcAntoineSchmidtQC, line 176: Not sure it's worth keeping the `_dense_matrix` function. We do not get the same clarity gains as before. What do you think @lbittarello ?

lbittarello: I don't think it's worth keeping either of them. 👍

src/tabmat/constructor.py
```diff
@@ -220,99 +255,83 @@ def from_polars(
...
-    if sparse_columns:
-        matrices.append(_sparse_matrix(df, sparse_columns, dtype))
-        indices.append(sparse_indices)
-        is_cat.append(False)
+def from_polars(
+    df,
+    dtype: np.dtype = np.float64,
+    sparse_threshold: float = 0.1,
+    cat_threshold: int = 4,
+    object_as_cat: bool = False,
```
lbittarello, line 278: This argument is not documented.

It's also a bit misleading in that we only convert string columns (rather than object columns).

lbittarello (approved):

PR #384, opened 2024-09-02
Switch to pixi

stanmart:
Checklist

Similarly to [glum #802](https://github.com/Quantco/glum/pull/802), this migrates the dev installation to pixi. It's more or less the same, so see that PR for the discussion.

TODO:
- [x] Fix the typing issues that mypy finds now that it sees the dependencies.
- [x] Find minimal versions of the required packages (and update oldies environment accordingly_)
- [x] Make readthedocs work
- [x] Fix `benchmark-analyze` (although I don't think it is a pixi issue)
- [x] Update contributing instructions


stanmart:
Since submitting this PR the a new version of pixi has been released, which supports [passing the `no-build-isolation` flag to uv](https://github.com/prefix-dev/pixi/issues/1124#issuecomment-2327937986). Maybe it's time to get rid of the `postinstall` workaround? I gave it a quick test and works, but there are a couple of caveats:

* It makes package resolution somewhat slower (I think it makes sure that the dependencies for `tabmat` are installed).
* I have not yet figured out how to manually trigger a rebuild of the package, which is necessary when the C++/Cython modules are changed.

What do you think @pavelzw?

.github/workflows/ci.yml
```diff
@@ -3,87 +3,57 @@ on: [push]
 
 jobs:
   pre-commit-checks:
-    name: pre-commit checks
+    name: Linux - pre-commit checks
     timeout-minutes: 30
     runs-on: ubuntu-latest
     env:
       PRE_COMMIT_USE_MICROMAMBA: 1
```
MarcAntoineSchmidtQC, line 10: I'm curious. I think this PR changes the way we run pre-commit hooks. Is PRE_COMMIT_USE_MICROMAMBA still necessary?

Glum also kept it, so maybe I'm misunderstanding this.

MarcAntoineSchmidtQC (approved):
I left 2 questions regarding the CI setup, but otherwise looks good. Tested on my mac and everything works perfectly.

MarcAntoineSchmidtQC:
Regarding the switch to the no-build-isolation using pixi, how slow does it make the package resolution? Is it going from like 10 seconds to 2 minutes, or from 10 seconds to 15 seconds? If it's the latter, then I'm absolutely okay with that.

For the rebuild, either we can keep the post-install task or rename it to "rebuild". Also, as this is only relevant for deep development, we can also instruct the contributors to remove the lock file and start from scratch.

stanmart:
> Is it going from like 10 seconds to 2 minutes, or from 10 seconds to 15 seconds?

If we make this change then [pixi has to install an environments](https://github.com/Quantco/copier-template-python-pixi/issues/18#issuecomment-2191073299) before being able to make the final solve. I did some testing: this means that from starting from scratch, environment resolution (e.g. when adding a new package) takes ~2 minutes on my laptop. Another issue is that these environments take up a lot of space (11 gigs). After all environments are installed, resolution is reasonably fast (e.g. adding a package takes around 10-15 seconds.

I'd say making devs install all environments when they want to add a dependency is too big of a tradeoff (especially in terms of disk space) for being able to omit postinstall. I would wait until pixi gets a `no-dependencies` flag for pypi dependencies, or some other solution.

> For the rebuild, either we can keep the post-install task or rename it to "rebuild". Also, as this is only relevant for deep development, we can also instruct the contributors to remove the lock file and start from scratch.

Good point! Even though I would not make the switch just yet, it would be a perfect solution when we do.

.github/workflows/ci.yml
stanmart: Good point. Probably not, as pre-commit is not installing anything anymore. I'll try if it works without this option and revert if it turns out to be necessary. :)

.github/workflows/ci.yml
```diff
@@ -3,87 +3,57 @@ on: [push]
...
-      - name: Run pytest (unix)
-        if: matrix.os != 'windows-latest'
-        shell: bash -el {0}
-        run: pytest -nauto tests -m "not high_memory" --doctest-modules src/
-      - name: Run pytest (windows)
-        if: matrix.os == 'windows-latest'
-        shell: powershell
-        run: pytest -nauto tests -m "not high_memory"
+        if: matrix.NOTE == 'Nightly Builds'
+        run: pixi run -e ${{ matrix.environment }} install-nightlies
```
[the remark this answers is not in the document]
MarcAntoineSchmidtQC: ohhh I don't know why I didn't see that the nightly was indeed part of the matrix above. Nevermind then.

.github/workflows/ci.yml
stanmart: Seems to work, so let's get rid of it.

pavelzw:
> Another issue is that these environments take up a lot of space (11 gigs)

xref https://github.com/prefix-dev/pixi/issues/1046

PR #389, opened 2024-09-12
Create CatMatrix from codes and categories

MarcAntoineSchmidtQC:
See [this issue](https://github.com/Quantco/glum/issues/838) on Glum to understand the reasoning.

Checklist

### DIFF
--- a/src/tabmat/categorical_matrix.py
+++ b/src/tabmat/categorical_matrix.py
@@ -324,6 +324,8 @@ class CategoricalMatrix(MatrixBase):
             indices = np.nan_to_num(cat_vec, nan=-1)
             if max(indices) >= len(categories):
                 raise ValueError("Indices exceed length of categories.")
+            if min(indices) < -1:
+                raise ValueError("Indices must be non-negative (or -1 for missing).")
         else:
             indices, self.categories = _extract_codes_and_categories(cat_vec)
 
@@ -380,13 +382,13 @@ class CategoricalMatrix(MatrixBase):
             "This property will be removed in the next major release.",
             category=DeprecationWarning,
         )
-
-        if _is_polars(self._input_dtype):
-            out = self.categories[self.indices].astype("object", copy=False)
-            out = np.where(self.indices < 0, None, out)
-            return pl.Series(out, dtype=pl.Enum(self.categories))
-
-        return pd.Categorical.from_codes(self.indices, categories=self.categories)
+        try:
+            return pd.Categorical.from_codes(self.indices, categories=self.categories)
+        except NameError:
+            raise ModuleNotFoundError(
+                "The `cat` property is provided for backward compatibility and "
+                "requires pandas to be installed."
+            )
 
     def recover_orig(self) -> np.ndarray:
         """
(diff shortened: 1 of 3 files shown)


MarcAntoineSchmidtQC:
@lbittarello, if you have time I would love to get your feedback. It removes the `_Categorical` class that you added during the polars PR.

stanmart (approved):
Looks good. I feel that a separate constructor, e.g. `CategoricalMatrix.from_codes` (like [`pandas.Categorical.from_codes`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Categorical.from_codes.html)) might be a bit clearer than making the meaning of `cat_vec` be dependant on another argument, but I'm on board with this solution, too.

lbittarello (commented):
I'm happy with the option to create a categorical matrix from codes. Indeed, it was my preferred option. :)

However, this PR has an unintended consequence. If we use a Polars array to instantiate a categorical matrix, the `cat` property returns a Polars array. If we then subset this categorical matrix, the subset will no longer remember that the original input was a Polars array, because the `_input_dtype` attribute will be set to the `dtype` of the array of categorical codes. At that point, `cat` will return a Pandas series.

The `_Categorical` container ensures that, no matter how many times we subset a categorical matrix, we'll always remember the input type and `cat` will always return an appropriate series.

I'm not sure that we care about it. I added a deprecation warning to `cat` to that we'd be able to take categorical codes in the future without these complications.

Alternatively, we could add an `input_dtype` instantiation parameter for `__getitem__` to set, but it seemed a strange attributes for users to have access to.

MarcAntoineSchmidtQC:
If we deprecate the `cat` property, I don't think we should be thinking about expanding its scope. Polars support is not released yet, so in the stable version `cat` is a `pandas.Categorical`. I would be happy to simply return a `pandas.Categorical` when possible and otherwise raise an error saying that this method is not supported with a non-pandas backend. This is fully backward compatible.

MarcAntoineSchmidtQC:
Of course, I can add a more obvious error message instead of relying on python spitting out an error because it doesn't know what `pd` is.

lbittarello:
> I would be happy to simply return a `pandas.Categorical` when possible and otherwise raise an error saying that this method is not supported with a non-pandas backend.

That's fine by me. :)

src/tabmat/categorical_matrix.py
```diff
@@ -321,13 +313,21 @@ def __init__(
...
         self._missing_category = cat_missing_name
 
-        indices, self.categories = _extract_codes_and_categories(cat_vec)
+        if categories is not None:
+            self.categories = categories
+            indices = np.nan_to_num(cat_vec, nan=-1)
+            if max(indices) >= len(categories):
+                raise ValueError("Indices exceed length of categories.")
+            if min(indices) < -1:
+                raise ValueError("Indices must be non-negative (or -1 for missing).")
```
lbittarello, line 328: Would it make sense to check that the indices are integers at this point?

MarcAntoineSchmidtQC: Absolutely.

MarcAntoineSchmidtQC: I added a try/except block surrounding the call to `.astype(np.int32)`. I don't want to check that the input dtype is an integer directly because I want to allow nan values.

lbittarello (approved):

PR #393, opened 2024-09-20
Check sandwich dimensions

stanmart:
Checklist

Add checks to raise an informative error on dimension or dtype mismatch in `MatrixBase.sandwich`. Before this, a dimension mismatch led to either `np.nan`s or simpoly wrong output depending on the platform.


MarcAntoineSchmidtQC (approved):
Thanks @stanmart! This is great.

MarcAntoineSchmidtQC:
Can you add a quick changelog entry? I think it's helpful to track when a previous misuse of a method now raises an error.

PR #423, opened 2024-12-04
Fix breakage caused by changes in formulaic

stanmart:
It seems like formulaic might be considering the `EncodedTermStructure` a private class based on the fact that it is being moved in a non-major release. A bigger refactor of the `formula` module might be worth considering at some point now that formulaic reached 1.0 and is more stable. Until then, this PR provides a quick fix for #422.

Checklist

### DIFF
--- a/src/tabmat/formula.py
+++ b/src/tabmat/formula.py
@@ -10,7 +10,6 @@ import pandas
 from formulaic import ModelMatrix, ModelSpec
 from formulaic.errors import FactorEncodingError
 from formulaic.materializers import FormulaMaterializer
-from formulaic.materializers.base import EncodedTermStructure
 from formulaic.materializers.types import FactorValues, NAAction, ScopedTerm
 from formulaic.parser.types import Term
 from formulaic.transforms import stateful_transform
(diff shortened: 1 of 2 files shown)


src/tabmat/formula.py
```diff
@@ -24,6 +23,11 @@
 from .sparse_matrix import SparseMatrix
 from .split_matrix import SplitMatrix
 
+try:
+    from formulaic.materializers.base import EncodedTermStructure
+except ImportError:
+    from formulaic.materializers.types.formula_materializer import EncodedTermStructure
```
lbittarello, line 29: Could we use the `formulaic` version to select the import path instead of a `try/except` block?

stanmart: Unfortunately I don't know in which version the change will appear. It's on main now, and my guess is `1.1`, but I'm not sure.

lbittarello (approved):
Maybe raise an issue with `formulaic` about breaking the public API without a major release?

MarcAntoineSchmidtQC:
@stanmart, I hijacked your PR to also update a CI runner so that the CI would pass.

MarcAntoineSchmidtQC:
oops, I thought it would be a simple fix. Not the case.

I will open a different PR to fix this.

MarcAntoineSchmidtQC (approved):

stanmart:
haha, all good, I'll merge it then

Issue #414, opened 2024-11-12
Standardize `DenseMatrix` by shifting data


mlondschien:
```python
In [1]: import numpy as np
   ...: import tabmat
   ...: 
   ...: n = 5_000
   ...: p = 1_000
   ...: 
   ...: rng = np.random.default_rng(0)
   ...: means = rng.exponential(10, p) ** 2
   ...: stds = rng.exponential(10, p) ** 2
   ...: 
   ...: X = rng.uniform(size=(n, p)) * stds + means
   ...: 
   ...: matrix = tabmat.DenseMatrix(X)
   ...: standardized_matrix1, emp_mean1, emp_std1 = matrix.standardize(np.ones(n) / n, True, True)
   ...: 
   ...: emp_mean2 = X.mean(axis=0)
   ...: emp_std2 = X.std(axis=0)
   ...: X = (X - emp_mean2) / emp_std2
   ...: standardized_matrix2 = tabmat.DenseMatrix(X)
   ...: 
   ...: weights = rng.uniform(size=n)

In [2]: %%timeit
   ...: sandwich1 = standardized_matrix1.sandwich(weights)
   ...: 
   ...: 
50.9 ms ± 2.72 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [3]: %%timeit
   ...: sandwich2 = standardized_matrix2.sandwich(weights)
   ...: 
   ...: 
34.2 ms ± 602 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [4]: %%timeit
   ...: sandwich3 = X.T @ np.diag(weights) @ X
   ...: 
   ...: 
352 ms ± 24.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [5]: sandwich1 = standardized_matrix1.sandwich(weights)
   ...: sandwich2 = standardized_matrix2.sandwich(weights)
   ...: sandwich3 = X.T @ np.diag(weights) @ X
   ...: 
   ...: print(np.max(np.abs(sandwich1 - sandwich2)))
   ...: print(np.max(np.abs(sandwich1 - sandwich3)))
   ...: print(np.max(np.abs(sandwich2 - sandwich3)))
0.06973587287713845
0.0697358728771389
8.881784197001252e-16
In [6]: print(np.max(np.abs(sandwich1.T - sandwich1)))
0.0
```
[...]

PR #436, opened 2025-01-24
More robust `DenseMatrix._get_col_stds`

mlondschien:
xref #414 
Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,8 @@ jobs:
         run: pixi run -e ${{ matrix.environment }} install-nightlies
       - name: Install repository
         run: pixi run -e ${{ matrix.environment }} postinstall
+      - name: Run Malte's example
+        run: pixi run -e ${{ matrix.environment }} python example.py 
       - name: Run pytest
         run: pixi run -e ${{ matrix.environment }} test -nauto -m "not high_memory"
       - name: Run doctest
--- a/example.py
+++ b/example.py
@@ -1,0 +1,37 @@
+import numpy as np
+
+import tabmat
+
+np.set_printoptions(suppress=True)
+
+for dtype in [np.float32, np.float64]:
+    X = np.array(
+        [
+            [46.231056, 126.05263, 144.46439],
+            [46.231224, 128.66818, 0.7667693],
+            [46.231186, 104.97506, 193.8872],
+            [46.230835, 130.10156, 143.88954],
+            [46.230896, 116.76007, 7.5629334],
+        ],
+        dtype=dtype,
+    )
+    v = np.array(
+        [0.12428328, 0.67062443, 0.6471895, 0.6153851, 0.38367754], dtype=dtype
+    )
+
+    weights = np.full(X.shape[0], 1 / X.shape[0], dtype=dtype)
+
+    stmat, out_means, col_stds = tabmat.DenseMatrix(X).standardize(weights, True, True)
+
+    print(stmat.toarray().T @ v)
+    print(stmat.transpose_matvec(v))
+
+    # compute by hand
+    res = np.zeros(X.shape[1], dtype=dtype)
+    for col in range(X.shape[1]):
+        res[col] += (
+            stmat.shift[col] + stmat.mult[col] * stmat.mat.toarray()[:, col]
+        ) @ v
+
+    print(res)
+    print("\n")
--- a/src/tabmat/dense_matrix.py
+++ b/src/tabmat/dense_matrix.py
@@ -164,8 +164,8 @@ class DenseMatrix(MatrixBase):
         raise TypeError
 
     def _get_col_stds(self, weights: np.ndarray, col_means: np.ndarray) -> np.ndarray:
-        """Get standard deviations of columns."""
-        sqrt_arg = transpose_square_dot_weights(self._array, weights) - col_means**2
+        """Get standard deviations of columns using weights `weights`."""
+        sqrt_arg = transpose_square_dot_weights(self._array - col_means[np.newaxis, :], weights)
         # Minor floating point errors above can result in a very slightly
         # negative sqrt_arg (e.g. -5e-16). We just set those values equal to
         # zero.
--- a/src/tabmat/dense_matrix.py
+++ b/src/tabmat/dense_matrix.py
@@ -165,7 +165,9 @@ class DenseMatrix(MatrixBase):
 
     def _get_col_stds(self, weights: np.ndarray, col_means: np.ndarray) -> np.ndarray:
         """Get standard deviations of columns using weights `weights`."""
-        sqrt_arg = transpose_square_dot_weights(self._array - col_means[np.newaxis, :], weights)
+        sqrt_arg = transpose_square_dot_weights(
+            self._array - col_means[np.newaxis, :], weights
+        )
         # Minor floating point errors above can result in a very slightly
         # negative sqrt_arg (e.g. -5e-16). We just set those values equal to
         # zero.
--- a/src/tabmat/standardized_mat.py
+++ b/src/tabmat/standardized_mat.py
@@ -130,7 +130,15 @@ class StandardizedMatrix:
         if not hasattr(d, "dtype"):
             d = np.asarray(d)
         check_sandwich_compatible(self, d)
-
+        # stat_mat = mat * mult[newaxis, :] + shift[newaxis, :]
+        # stat_mat.T @ d[:, newaxis] * stat_mat
+        # = mult[:, newaxis] * mat.T @ d[:, newaxis] * mat * mult[newaxis, :] +   (1)
+        #   mult[:, newaxis] * mat.T @ d[:, newaxis] * np.outer(ones, shift) +    (2)
+        #   shift[:, newaxis] @ d[:, newaxis] * mat * mult[newaxis, :] +          (3)
+        #   shift[:, newaxis] @ d[:, newaxis] * shift[newaxis, :]                 (4)
+        #
+        # (1) = self.mat.sandwich(d) * np.outer(limited_mult, limited_mult)
+        # (2) = mult * self.transpose_matvec(d) * shift[newaxis, :]
         if rows is not None or cols is not None:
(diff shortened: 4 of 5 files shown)


mlondschien:
A non-zero constant column results in
```
sqrt_arg = transpose_square_dot_weights(self._array, weights) - col_means**2
```
close to machine precision ~1e-15. Then `_get_col_stds` returns ~1e-8 for that column. The resulting `mult` of the `StandardizedMatrix` is large ~1e8, and so at least one entry of `np.outer(mult, mult)` blows up to `1e15`. As a result, the corresponding entry in `StandardizedMatrix.sandwich(np.ones(n)/n)` no longer cancels out.

If `DenseMatrix._get_col_stds` returned a value close to machine precision, the `mult` is 1 and so these instabilities don't occur.

It might be worthwile to put a check somewhere in the init of `StandardizedMatrix`: If `mult` is close to `sqrt(machine_precision)`, the results of `sandwich` will be inaccurate.

src/tabmat/dense_matrix.py
```diff
@@ -164,8 +164,10 @@ def _cross_sandwich(
         raise TypeError
 
     def _get_col_stds(self, weights: np.ndarray, col_means: np.ndarray) -> np.ndarray:
-        """Get standard deviations of columns."""
-        sqrt_arg = transpose_square_dot_weights(self._array, weights) - col_means**2
+        """Get standard deviations of columns using weights `weights`."""
+        sqrt_arg = transpose_square_dot_weights(
+            self._array - col_means[np.newaxis, :], weights
```
stanmart, line 169: Can we do this within the loop in `transpose_square_dot_weights` so that we can avoid (temporarily) allocating a new array that's as large as `self._array`?

example.py
```diff
@@ -0,0 +1,37 @@
+import numpy as np
```
stanmart, line 1: It would be nice to turn this into a test case so that we don't reintroduce the issue accidentally.

stanmart:
Nice find, thank you!

src/tabmat/dense_matrix.py
stanmart: It's a private function that we don't use anywhere else I believe, so also feel free to change its arguments (i.e., pass `col_means` to it) if that feels most natural.

stanmart:
Would the same operation for the sparse matrix also benefit from a similar change? Hopefully we have much fewer non-zero constant columns there, but still :)

https://github.com/Quantco/tabmat/blob/866257162d619c925a3d0772ea1b31c4de87dfe8/src/tabmat/sparse_matrix.py#L297-L306

mlondschien:
Am I seeing this correctly and no tests are run as part of ci?

mlondschien:
> Would the same operation for the sparse matrix also benefit from a similar change? Hopefully we have much fewer non-zero constant columns there, but still :)
> 
> https://github.com/Quantco/tabmat/blob/866257162d619c925a3d0772ea1b31c4de87dfe8/src/tabmat/sparse_matrix.py#L297-L306

I guess this would be a more complex change if we want to avoid a loop through all (including zero) entries.

stanmart:
> Am I seeing this correctly and no tests are run as part of ci?

We should probably change our CI config such that tests run on pull requests, and not just on pushes.

> I guess this would be a more complex change if we want to avoid a loop through all (including zero) entries.

Good point. Let's just skip it then, it should not be an in issue anyways.

.github/workflows/ci.yml
```diff
@@ -48,6 +53,8 @@ jobs:
         run: pixi run -e ${{ matrix.environment }} install-nightlies
       - name: Install repository
         run: pixi run -e ${{ matrix.environment }} postinstall
+      - name: Run Malte's example
+        run: pixi run -e ${{ matrix.environment }} python example.py
```
stanmart, line 57:
```suggestion
```
No need to run the example in the CI

example.py
stanmart, line 1: Please remove this file before merging. Something like this could be a nice test case, though.

src/tabmat/standardized_mat.py
```diff
@@ -130,7 +130,15 @@ def sandwich(
...
-
+        # stat_mat = mat * mult[newaxis, :] + shift[newaxis, :]
+        # stat_mat.T @ d[:, newaxis] * stat_mat
+        # = mult[:, newaxis] * mat.T @ d[:, newaxis] * mat * mult[newaxis, :] +   (1)
+        #   mult[:, newaxis] * mat.T @ d[:, newaxis] * np.outer(ones, shift) +    (2)
+        #   shift[:, newaxis] @ d[:, newaxis] * mat * mult[newaxis, :] +          (3)
+        #   shift[:, newaxis] @ d[:, newaxis] * shift[newaxis, :]                 (4)
+        #
+        # (1) = self.mat.sandwich(d) * np.outer(limited_mult, limited_mult)
+        # (2) = mult * self.transpose_matvec(d) * shift[newaxis, :]
```
stanmart, line 141: Is this comment meant to be added to the code? In that case, I could use some context here 😅

stanmart (approved):
I will run the benchmarks in a bit to make sure there is no performance regression due to these changes, but otherwise LGTM.

.github/workflows/ci.yml
```diff
@@ -1,5 +1,10 @@
 name: CI
-on: [push]
+
+on:
+  pull_request:
+    branches:
+      - main
+  push:
```
stanmart, line 7: Maybe let's do

```suggestion
on:
  # We would like to trigger for CI for any pull request action -
  # both from QuantCo's branches as well as forks.
  pull_request:
  # In addition to pull requests, we want to run CI for pushes
  # to the main branch and tags.
  push:
    branches:
      - "main"
    tags:
      - "*"
```

instead to be consistent with [glum](https://github.com/Quantco/glum/blob/b471522c611b263c00ae841fd0f46660c31a6d5f/.github/workflows/ci.yml#L2-L12) and [metalearners](https://github.com/Quantco/metalearners/blob/250a0c01a1d1442488a07c90bb55540931336ad8/.github/workflows/ci.yml#L2-L12). I believe this avoids running the CI twice when a PR is submitted from a tabmat branch to main.

src/tabmat/standardized_mat.py
mlondschien: This is similar to the comment in `StandardizedMatrix.transpose_matvec`. I needed to write this down. We can also remove this.

example.py
stanmart: I believe he just added it to illustrate the issue in a reproducible linux-x86 environment, not to actually merge into main. But we can wait for him to chime in on this.

> Do you want to add this as a test?

Yes, a simplified version of this, or your examples from the related issue could make a great test case IMO.

mlondschien:
Any change we can merge this and do a release :)? I can remove Jan's file and add this as a test. Where should I add it to?

stanmart:
Yes! I can add the test if you don't mind me pushing to this branch. We'll probably want to add a changelog entry, too. How about something like "Improved the robustness of standardization for dense matrices with constant columns"?

mlondschien:
Feel free to push as you'd like.

I already added

> **Bug fix:**

> - A more robust :meth:`DenseMatrix._get_col_stds` results in more accurate :meth:`StandardizedMatrix.sandwich` results.

stanmart:
Ah, I can't because it's your repo. It's fine though. So I thought something like this might be good in `tests/test_matrices.py`:

```python
@pytest.mark.parametrize("dtype", [np.float32, np.float64])
def test_standardize_constant_cols(dtype):
    X = np.array(
        [
            [46.231056, 126.05263, 144.46439],
            [46.231224, 128.66818, 0.7667693],
            [46.231186, 104.97506, 193.8872],
            [46.230835, 130.10156, 143.88954],
            [46.230896, 116.76007, 7.5629334],
        ],
        dtype=dtype,
    )
    v = np.array(
        [0.12428328, 0.67062443, 0.6471895, 0.6153851, 0.38367754], dtype=dtype
    )
    weights = np.full(X.shape[0], 1 / X.shape[0], dtype=dtype)

    standardized_mat, _, _ = tm.DenseMatrix(X).standardize(
        weights, center_predictors=True, scale_predictors=True
    )

    result = standardized_mat.transpose_matvec(v)
    expected = standardized_mat.toarray().T @ v

    np.testing.assert_allclose(result, expected)
```

The issue is that the test still fails for `float32`. Or is that expected and this branch does not fully fix what `example.py` demonstrates?

mlondschien:
image
maybe you are not a maintainer 🤷

stanmart:
Haha, I thought I was, but who knows 😅

mlondschien:
[c02fbc2](https://github.com/Quantco/tabmat/pull/436/commits/c02fbc2063e2be25ba76e8a94633739fc31908cb) still fails.

stanmart:
Yeah, so it's failing for `float32` still. Maybe we shoud add a more direct test instead? It's the standard deviation calculation which was improved after all

For example this:

```python
@pytest.mark.parametrize("dtype", [np.float32, np.float64])
def test_standardize_almost_constant_cols(dtype):
    X = np.array(
        [
            [46.231056, 126.05263, 144.46439],
            [46.231224, 128.66818, 0.7667693],
            [46.231186, 104.97506, 193.8872],
            [46.230835, 130.10156, 143.88954],
            [46.230896, 116.76007, 7.5629334],
        ],
        dtype=dtype,
    )
    v = np.array(
        [0.12428328, 0.67062443, 0.6471895, 0.6153851, 0.38367754], dtype=dtype
    )
    weights = np.full(X.shape[0], 1 / X.shape[0], dtype=dtype)

    _, means, stds = tm.DenseMatrix(X).standardize(
        weights, center_predictors=True, scale_predictors=True
    )

    decimal = 3 if dtype == np.float32 else 6
    np.testing.assert_almost_equal(means, X.mean(axis=0), decimal=decimal)
    np.testing.assert_almost_equal(stds, X.std(axis=0), decimal=decimal)
```
fails on main and passes on this branch which is what we want to see.

stanmart:
Let me try again if I can make changes to this branch

stanmart (approved):

PR #439, opened 2025-01-30
Bump QEMU and release 4.1.1 again

stanmart:
Yesterday's 4.1.1 release [did not go well](https://github.com/Quantco/tabmat/actions/runs/13031194083), as there were a bunch of segfaults while building jemalloc for linux-aarch64. As it turns, out, we have been using a rather old QEMU version, because the `setup-qemu-action` pulls the image with the `latest` tag. It seems sensible enough, but as it turns out, this tag [has not been updated](https://github.com/tonistiigi/binfmt/issues/165) for the relevant container in more than two years.

Switching to an up-to-date QEMU seems to [solve the problem](https://github.com/Quantco/tabmat/actions/runs/13039638199/job/36378339147), Therefore I propose upgrading to this recent QEMU version and attempting a new release today.

Checklist


stanmart:
fyi @mlondschien

jtilly (approved):

mlondschien:
thank you!

PR #452, opened 2025-03-20
Fix column validation when all levels of a categorical are dropped while ensuring full rank

stanmart:
This PR fixes [glum #918](https://github.com/Quantco/glum/issues/918). The cause of the issue was that we have to do a bit of work for `formulaic` to be able to validate the generated columns when generating the model matrix from an already existing `ModelSpec`, and we missed the case where a categorical variable would have zero categories.

Checklist
* [X] Added a `CHANGELOG.rst` entry

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,8 @@ Changelog
 
 **Bug fix:**
 
-- Fixed a bug which incorrectly raised an error when constructing tabmat matrices from existing ``ModelSpec``s when they contained a categorical column without any levels.
+- Fixed a bug which caused issues when constructing tabmat matrices from existing ``ModelSpec``\s when they contained categorical columns with all levels dropped.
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -7,6 +7,13 @@
 Changelog
 =========
 
+4.1.2 - UNRELEASED
+------------------
+
+**Bug fix:**
+
+- Fixed a bug which incorrectly raised an error when constructing tabmat matrices from existing ``ModelSpec``s when they contained a categorical column without any levels.
```
lbittarello, line 15:
```suggestion
- Fixed a bug which incorrectly raised an error when constructing tabmat matrices from existing ``ModelSpec``\s when they contained a categorical column without any levels.

```

lbittarello (approved):

PR #489, opened 2025-10-20
Fix transpose_matvec for categorical matrix with read-only indices

paolamalsot:
The bug would lead to "buffer source array is read-only" when calling `transpose_matvec` on a categorical matrix with a large number of categories (see`test_transpose_matvec.py`).

Indeed, during the initialization of a categorical matrix with a high number of categories, the line `self.indices = indices.astype(np.int32, copy=False)` in the constructor would not create a copy, and therefore `self.indices` would be a read-only buffer. As a result, the Cython code in `transpose_matvec_complex` and `transpose_matvec_fast` failed.

In the fix, the underlying Cython functions `transpose_matvec_fast` and `transpose_matvec_complex` allow for read-only buffers in their input `indices`.

Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@
 Changelog
 =========
 
+4.1.4 - 2025-10-20
+------------------
+
+**Bug fix:**
+
+- Fixed :meth:`CategoricalMatrix.transpose_matvec` to operate on read-only buffers as well.
+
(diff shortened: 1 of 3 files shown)


paolamalsot:
I think I ended up trying this because of compilation bugs but this simpler solution seems to work as well ! Thanks !

PR #537, opened 2026-09-04
Respect sparse_threshold for pandas SparseDtype columns

### LINKED ISSUE #378: Improve performance of `tabmat.from_pandas` for sparse columns
Instead of

https://github.com/Quantco/tabmat/blob/c0c8626b0b35737e34742c422f6adc41d0b8001f/src/tabmat/constructor.py#L158

we should consider

```python
if isinstance(coldata.dtype, pd.SparseDtype):
    sparse_density = coldata.sparse.density
else:
    sparse_density = (coldata != 0).mean()

if sparse_density <= sparse_threshold:
    ...
```

[pandas.Series.sparse.density](https://pandas.pydata.org/docs/reference/api/pandas.Series.sparse.density.html) operates in constant time (e.g., 3e-5s on my machine) while [pandas.Series.ne](https://pandas.pydata.org/docs/reference/api/pandas.Series.ne.html) and [pandas.Series.mean](https://pandas.pydata.org/docs/reference/api/pandas.Series.mean.html) are linear in the amount of non-zero entries for sparse columns (e.g., 3e-3s on my machine for one million rows with density 0.1).

The proposed changes reduce the time to convert a `pandas.DataFrame` with 130 sparse columns (and a few categoricals) and one million rows from 0.5s to 0.15s.

Danila-Pechenev:
Fixes #378.

## Summary

Update `from_df` so pandas `SparseDtype` columns respect `sparse_threshold`
instead of being unconditionally routed to sparse storage.

## Why

Issue #378 predates the current Narwhals-based `from_df` implementation.
In current `main`, pandas sparse columns are handled separately because their
dtype is not supported by Narwhals, but this special case bypasses
`sparse_threshold`.

The updated logic preserves logical non-zero density semantics while avoiding
unnecessary scans for low-density zero-fill sparse columns.

For zero-fill `SparseDtype` columns, pandas sparse storage density is used as a
fast first check. If it is already at or below the threshold, the column is
known to be sparse without scanning its logical values. If it is above the
threshold, logical density is evaluated so explicitly stored zeros are handled
correctly.

For non-zero fill values, classification uses logical non-zero density because
pandas storage density does not represent non-zero density in that case.

Empty and missing-fill sparse columns preserve the existing sparse behavior.

## Tests

Added regression coverage for:

- zero-fill sparse columns below and above `sparse_threshold`;
- density exactly at the threshold;
- explicitly stored zeros;
- non-zero fill values where storage and logical densities disagree;
- empty pandas sparse columns;
- missing fill values.

## Checklist


MarcAntoineSchmidtQC:
When we wrote this function, the intention was that sparse_threshold would be used to sparsify columns and not densify already sparse ones. The documentation is not clear on this, so I understand why this PR was made.

In my mind, users that pass in sparse columns already have checked that it made sense for their column to be sparse, so we don't want to overwrite this.

When looking at #378, the issue had actually been fixed by moving to narwhal. Bypassing the density check to keep the sparse columns sparse solved the inefficiency.

I see 3 questions/observations from this PR:

1. Should we densify an already sparse column that is below the sparse_threshold?
2. Tabmat's SparseMatrix is intended to work with zero-filled sparse arrays and nothing else. This PR made me realize that we don't deal with those cases. So we should check that the fill value is always zero.
3. The SparseMatrix creation itself is quite inefficient right now when the input data is a pandas SparseArray. Should we improve this? (not addressed in this PR but related).

@stanmart, I would love to get your input here. What do you think?

stanmart:
I agree. I would avoid densifying user-supplied sparse columns as it might be surprising behavior. Improving the related documentation would be a nice addition, though.

2 and 3 are also good points. 2 in particular should be a simple fix and is definitely worth doing. If there is a good way to improve the performance of `SparseMatrix` creation, that's great as well.

Danila-Pechenev:
Thanks, that makes sense. I had interpreted `sparse_threshold` as controlling the representation of all numeric columns, rather than only sparsifying columns that are not already sparse.

I see that #542 now addresses the sparse construction and fill-value issues from this discussion. Since the intended one-way semantics of `sparse_threshold` are not currently clear from the documentation, I can narrow this PR down to a documentation clarification if that would be useful.

Would you prefer that I do that in this PR, or close this one and open a separate documentation PR?

stanmart:
> Since the intended one-way semantics of sparse_threshold are not currently clear from the documentation, I can narrow this PR down to a documentation clarification if that would be useful.

Yes, that would be amazing, thank you!

> Would you prefer that I do that in this PR, or close this one and open a separate documentation PR?

Up to you. Both are fine with me.

Danila-Pechenev:
Updated the PR to a documentation-only clarification as discussed.

stanmart (approved):
Thank you!

PR #541, opened 2026-09-15
Fix benchmark harness recording zero timings under pandas 3

abelianbee:
The benchmark harness writes results with `times["time"].iloc[i] = ...`, which is a chained assignment. Under pandas 3 copy-on-write that is rejected with `ChainedAssignmentError` and the value is dropped, so every timing and memory figure the harness produced was its initial value. Switched to `.loc`.

The new test fails on `main` and passes with the fix.

Checklist


tests/test_benchmark_cli.py
```diff
@@ -49,3 +51,25 @@ def test_run_all_benchmarks(
...
+        matrices,
+        include_baseline=False,
+        name="dense",
+        standardized=False,
+        ops_to_run=["matvec", "transpose-matvec", "sandwich"],
+        n_iterations=1,
+        bench_memory=False,
+    )
+    assert times["time"].notna().all()
+    assert (times["time"] > 0).all()
```
stanmart, line 75:
```suggestion
    assert times["memory"].notna().all()
```

Nit: this is a slightly more direct test

stanmart (approved):
Thank you, tiny nit aside looks good.

abelianbee:
Is this waiting on anything from me, or just a merge?

stanmart:
I had a tiny suggestion [here](https://github.com/Quantco/tabmat/pull/541#discussion_r4015195010) that you can apply, otherwise just the CI (started it just now) and the merge.

abelianbee:
Thank you!

stanmart:
Thank you for the fix! Merging rn

PR #542, opened 2026-09-15
Build the sparse block of from_pandas directly from pandas sparse columns

abelianbee:
Follows up on points 2 and 3 of the discussion in #537. xref #378.

`from_pandas` built its sparse block with `sps.coo_matrix(df[:, sparse_dfidx])`, which materializes the whole block densely before converting. For pandas sparse columns that is O(n * k) time and memory on data that is O(nnz). This assembles the CSC from each column's `sp_index` and `sp_values` instead, and calls `eliminate_zeros()` since pandas sparse arrays can hold explicit zeros.

`from_pandas` end to end, median of 5 after a warmup, M1 Max:

| n | p | density | before | after | peak heap before | after |
|---:|---:|---:|---:|---:|---:|---:|
| 500,000 | 50 | 0.01 | 0.110 s | 0.006 s | 208 MB | 3.0 MB |
| 1,000,000 | 50 | 0.005 | 0.236 s | 0.006 s | 416 MB | 3.0 MB |

Both shapes have the same nnz and now take the same time.

On the fill value: `SparseMatrix` only represents a fill of 0, and nothing checked that, so a sparse column with another fill went into sparse storage with every row explicitly stored. Those columns are logically dense and now go to dense storage with a warning. I went with a warning rather than an error because pandas' default fill for a float sparse array is NaN, so `.astype("Sparse")` on a float column would otherwise start raising. Happy to make it an error if you would rather.

Checklist


src/tabmat/constructor.py
```diff
@@ -105,9 +106,24 @@ def from_df(
...
+            if isinstance(native_dtype, pd.SparseDtype):
+                fill = native_dtype.fill_value
+                if np.isscalar(fill) and not pd.isna(fill) and fill == 0:
+                    pandas_sparse_dfidx.add(dfcolidx)
+                    sparse_dfidx.append(dfcolidx)
+                    sparse_tmidx.append(mxcolidx)
+                else:
+                    # SparseMatrix only represents zero fill. A sparse array with
+                    # any other fill value is logically dense, so store it that way.
+                    warnings.warn(
```
stanmart, line 119: It might be slightly more elegant to collect all non-sparsified columns first and emit a single warning that lists them instead of possibly dozens of separate warnings.

src/tabmat/constructor.py
```diff
@@ -190,9 +206,38 @@ def from_df(
...
+        # Assemble the CSC column by column. Going through
+        # ``sps.coo_matrix(df[:, sparse_dfidx])`` would materialize the whole block
+        # densely first, which for pandas sparse columns is O(n * k) work and memory
+        # on data that is O(nnz).
+        data_parts: list[np.ndarray] = []
+        index_parts: list[np.ndarray] = []
+        indptr = [0]
+        for dfcolidx in sparse_dfidx:
+            if dfcolidx in pandas_sparse_dfidx:
+                sparse_array = nw.to_native(df[:, [dfcolidx]]).iloc[:, 0].array
```
stanmart, line 218:
```suggestion
                sparse_array = nw.to_native(df[:, dfcolidx]).array
```

nit: `df[:, dfcolidx]` is already a (narwhals) series

CHANGELOG.rst
```diff
@@ -12,6 +12,7 @@ Changelog
 
 **Other changes:**
 
+- :func:`from_pandas` now builds its sparse block directly from pandas sparse columns instead of densifying them first, which makes construction much faster and uses memory proportional to the number of stored values rather than to the full block. Sparse columns whose fill value is not 0 are now stored as dense with a warning, since :class:`SparseMatrix` only represents a fill value of 0.
```
stanmart, line 15:
```suggestion
- :func:`from_df` now builds its sparse block directly from pandas sparse columns instead of densifying them first, which makes construction much faster and uses memory proportional to the number of stored values rather than to the full block. Sparse columns whose fill value is not 0 are now stored as dense with a warning, since :class:`SparseMatrix` only represents a fill value of 0.
```

Let's refer to the non-deprecated version

stanmart (commented):
Small comments aside looks great! Thank you, very valuable improvement.

abelianbee:
Thanks for the review, all three are in.

Batched the warning: the non-zero-fill columns are collected during the column loop and reported in one warning after it, the same way `ignored_cols` already works. The message lists each column with its fill value. Added a test that three such columns produce exactly one warning.

Took the other two suggestions as-is.

stanmart (approved):
Thank you for the follow-up, lgtm! Let's wait for CI but we can merge as soon as it's green.

abelianbee:
Pushed a fix for the nightly failure. It was the test, not the change: it built its input by multiplying two sparse arrays to get explicitly stored zeros, and newer pandas prunes those, so the setup assertion failed. It now builds the column with `SparseArray.from_spmatrix` from a scipy matrix that carries a stored zero, which is deterministic across versions, and additionally checks that the genuine non-zeros survive.

The new run needs workflow approval before it can go green.

stanmart:
There is a trivial conflict in the changelog due to merging the other PR, but looks good!

abelianbee:
Rebased on main, both changelog entries kept.

stanmart:
CI failure is unrelated to the PR, it's just a flaky micromamba download. Rerunning it rn

